Issue #33, opened 2020-05-09
make the input of flower demo to raw image data rather than file path


hanxiao:
to make it more realistic in practice, e.g. user can upload an image to search for similar flower.

PR #37, opened 2020-05-13
fix(flower-search): fixed issue #33

iego2017:
fix(flower-search): fixed issue #33


iego2017:
updated README.md in PR #40.

PR #50, opened 2020-05-15
fix: fix the broken flows when gateway pods are shared

### LINKED ISSUE #49: Southpark-search indexing process got stuck
I have been trying to run southpark-search following the instruction and the process seemed to stuck somewhere every time. (Before I have already tried `jina hello-world` and it can be executed without a problem.)

At first I thought it's still in progress just very slowly. But then I checked the system monitor they all seem kinda "idle". Nothing changed even after 20+ minutes. 

## Environment

Desktop: Fedora 31 (kernel 5.6)
Python version: 3.7.7
Jina version: 0.1.10 (installed via PyPi)

## Steps

- Cloned the project to my local folder
- `pip install -r requirements.txt` **> Success**
- `bash ./get_data.sh` **> Success**
- `python app.py -t index -n 10000` **> Process Stuck**

## Logs
Below is the logs that I reproduced today:
https://gist.github.com/pswu11/0724531689b5f700a40ccd4384c3fb15

## Screenshot
Screenshot from 2020-05-14 12-06-47

nan-wang:
resolves #49 

fix the misusage of the `gateway` Pod. Under the hood, the `gateway` Pod uses `CONNECT` to get connected with the other Pods while the normal Pods use `BIND`. The `CONNECT` is not stable when more than one Pods get connected to the same one. Especially, when the two pathways diverge from the `gateway` and later merged at the `join`, this leads to troubles. When one message is lost in one pathway, the whole Flow would get stuck.

### DIFF
--- a/news-search/flow-index.yml
+++ b/news-search/flow-index.yml
@@ -1,12 +1,16 @@
 !Flow
 pods:
+  dispatcher:
+    yaml_path: _forward
+    replicas: 2
+
   doc_indexer:
     yaml_path: pods/doc_indexer/doc_indexer.yml
     replicas: 2
 
   extractor:
     yaml_path: pods/extractor/extractor.yml
-    needs: gateway
+    needs: dispatcher
     replicas: 2
 
   encoder:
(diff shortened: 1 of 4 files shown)


hanxiao:
?? let's discuss this tomorrow.

PR #57, opened 2020-05-26
Rename example

YueLiu-jina:
add --CHN to 2 examples' folder name


YueLiu-jina:
Add --CHN to two examples' folder name

YueLiu-jina:
> 1. the language code for chinese is `zh`, not CHN
> 2. simply create a new directory `zh`, and put these two chinese tutorials there.

PR #59, opened 2020-05-26
Rename examples

YueLiu-jina:


YueLiu-jina:
folder renamed to zh

hanxiao:
lgtm, but let's hope there is no absolute path used in the docs, otherwise changing the directory will break those links. You may need to double check.

YueLiu-jina:
Get it! I'll do it tomorrow.

PR #79, opened 2020-06-25
feat(urbandict): add rest-api for the urbandict search example

YueLiu-jina:


YueLiu-jina:
This also has two problems remaining:

1.Same with flower search, docker push and update readme haven't been done. 
2.The dataset in this example should be downloaded from this link https://www.kaggle.com/therohk/urban-dictionary-words-dataset/download. It can only be done manually, curl/wget didn't work. So I wrap the dataset(200+Mb) into the docker image. But if others want to rebuild the image, they have to first download the dataset into the rest-api folder. This could be a troublesome process for them

Issue #74, opened 2020-06-18
[UrbanDict] ruamel.yaml.constructor.ConstructorError: could not determine a constructor for the tag '!DocKVIndexDriver'


xubiuit:
Hi guys,

I am trying to setup the UrbanDict example.
Instead using index 10000, I just used 100 (since I just want to test a bit, and check out the process only).
But I faced this when running `python app.py -t index -n 100`.
Env: `python3.8`, `jina-0.2.8`

```python
chunk_indexer@6939[I]:input tcp://0.0.0.0:51377 (PULL_BIND) 	 output tcp://0.0.0.0:51385 (PUSH_CONNECT)	 control over tcp://0.0.0.0:51379 (PAIR_BIND)
  chunk_indexer@6939[S]:ready and listening
    doc_indexer@6940[C]:unknown exception: ConstructorError(None, None, "could not determine a constructor for the tag '!DocKVIndexDriver'", <_ruamel_yaml.Mark object at 0x10f9ea770>)
Traceback (most recent call last):
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/jina/peapods/pea.py", line 343, in run
    self.loop_body()
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/jina/peapods/pea.py", line 305, in loop_body
    self.load_executor()
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/jina/peapods/pea.py", line 198, in load_executor
    self.executor = BaseExecutor.load_config(self.args.yaml_path,
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/jina/executors/__init__.py", line 404, in load_config
    return yaml.load(tmp_s)
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/ruamel/yaml/main.py", line 343, in load
    return constructor.get_single_data()
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/ruamel/yaml/constructor.py", line 113, in get_single_data
    return self.construct_document(node)
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/ruamel/yaml/constructor.py", line 118, in construct_document
    data = self.construct_object(node)
  File "/Users/binhnguyen/env/jina-python3-env/lib/python3.8/site-packages/ruamel/yaml/constructor.py", line 146, in construct_object
[...]

PR #83, opened 2020-06-30
feat(urbandict): update to new ingest api and add dockerfile

YueLiu-jina:
Our original code is not working due to the problem of driver !DocKVIndexDriver in yaml/index-doc.yml, as mentioned in #74 

I changed the code according to the southpark example and build a docker image without rest-api. Everything tested.


urbandict-search/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM pytorch/pytorch:latest
+
+WORKDIR /
+
+ENV DATA_DIR /
```
nan-wang, line 5:
```suggestion
ENV DATA_PATH /tmp/urban-dictionary-words-dataset.zip
```

urbandict-search/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+COPY . /
+
```
nan-wang, line 8: The data file should be copied into the docker so that `prepare_data.py` can be executed properly.
```suggestion
COPY urban-dictionary-words-dataset.zip ${DATA_PATH}
```

urbandict-search/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN python -m pip install --no-cache-dir --upgrade pip && \
+    pip install --no-cache-dir -r requirements_docker.txt
+
+RUN python -c "from transformers import DistilBertModel, DistilBertTokenizer; x='distilbert-base-cased'; DistilBertModel.from_pretrained(x); DistilBertTokenizer.from_pretrained(x)"
+
+RUN python prepare_data.py && python app.py -t index -n 10000
+
```
nan-wang, line 15: After building the index, delete the data file in order to keep the docker small.

```suggestion
RUN rm ${DATA_PATH}
```

urbandict-search/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+CMD ["python", "app.py","-t","query"]
```
nan-wang, line 16: add a new mode for serving with RESTful APIs.
```suggestion
CMD ["python", "app.py", "-t", "query_restful"]
```

PR #126, opened 2020-07-31
fix(urbandict): add splitter

JoanFM:
Hey @YueLiu-jina .

This is the idea I had, it is now broken due to a problem in `transformers` when dealing with different lengths of tokens.

The idea is that here we have the `word` and the set of `definitions`. The word does not have to be `encoded`, therefore I add a splitter before the `sentencizer` to extract the word and set it as `meta_info`, so that later we can say. The closest word to the `definition given` is the one extracted from the `doc.matches[idx].meta_info.decode()`.

Does this sound like a good idea @nan-wang?

If the aim of the example is to search for `words` given a `definition`, the word itself should not be encoded.

Now it works if the versions required by https://github.com/jina-ai/jina/issues/753 are used.

I understand the flow. We are using the closest definition which is the single definition present in the first chunk of a `match`.

But what I do not understand is why the `match.text` is empty. It it missing in the `DocIndexer`, or just not extracted? @nan-wang any light on this?

Anyway. Now the `example` works with this change.

### DIFF
--- a/urbandict-search/app.py
+++ b/urbandict-search/app.py
@@ -27,8 +27,8 @@ def print_topk(resp, word):
             score = match.score.value
             if score <= 0.0:
                 continue
-            doc = match.chunks[0].text
-            word, word_def = doc.split('+-=', maxsplit=1)
+            word_def = match.chunks[0].text
+            word = match.meta_info.decode()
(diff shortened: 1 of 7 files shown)


YueLiu-jina:
This makes a lot of sense to me

PR #132, opened 2020-08-03
[Merge after  jina-ai/jina#851] fix: change flows

JoanFM:
This `Draft` is open to clarify a few issues on how this example should be implemented.

**Changes introduced**
I think that `urbandict` is not working as it is supposed.

These changes inser the flow I think it should follow. Although I have not been able to make it work b/c of problems at DocIndexer. I can't make it use `matches` as traverse key, and I do not know how to make it use `parent_id` as key for the Indexer query.

The idea can be summarized:
**Index time**

```bash
!Flow
pods:
  splitter:
    uses: yaml/craft-split.yml
    **explanation**: First, at doc-level-0, we extract the word and put it into **meta_info**. and keep the definitions as a comma-separated set of **definitions**. In the next step (**CompoundExecutor**), it uses a **Sentencizer** to generate from this `level-0-depth document`, a set of chunks (`level-1-depth documents`) (as many chunks as number of descriptions (; separated))
  encoder:
    **explanation**: Here, encode every description (every chunk (`level-1-depth document`)). This is why IndexRequest has depth_range: [1, 1]
  chunk_indexer:
    **explanation**: Put every embedding to a vector Index. As a CompundIndexer, keep a key-value index, with the key being the chunk id (`level-1-depth document`) and the value its PB content.
  doc_indexer:
    uses: yaml/index-doc.yml
    needs: gateway
    **explanation**: Take directly the document (`level 0`) and index its document. It would even be cooler to split `craft` into 2 pods, and have it need the `splitter` instead of the `gateway`. This way it could keep the text as the set of descriptions and the word in the meta-info (So it can be extracted like this at query time).
  join_all:
    uses: _merge
    needs: [doc_indexer, chunk_indexer]
    read_only: true
```

**Search time**

```bash
!Flow

pods:
  encoder:
    uses: yaml/encode.yml
    **explanation**: Since the input will be a simple description, no need to craft it. So we need to encode directly the level-0 document. This is why **SearchRequest** has depth_range: [0, 0]
  chunk_indexer:
    uses: yaml/index-chunk.yml
    **explanation**: Here it needs to search for the closest vectors in the **vector indexer**. It needs to search based on the 0-depth document. Afterwards, from the **matches** (that is why matches is used as traverse key), extract the values from the key-value index.
  ranker:
    uses: MinRanker
  doc_indexer:
    uses: yaml/index-doc.yml
    **explanation**: I need to traverse on `matches` and extract from key-value pair based on the parent_id of the matches. Does it make sense? We would need something like this (Although I think we need a different driver for it) (see PR link) 
```


nan-wang:
yes, it makes sense. One will need to traverse matches and find the parent_id

JoanFM:
And how one would implement this pattern with the actual codebase?

JoanFM:
So maybe we need to change in  jina-ai/jina#758 to have a new driver. If we extract from parent_id we do not want to merge with the document maybe, we just want to extract what is in there? or merge and then exclude some fields?

nan-wang:
I see. The point is that there is no way to track the parent id all the way back to the top level.

JoanFM:
Should work after issue https://github.com/jina-ai/jina/issues/763 is solved

JoanFM:
I guess after closing https://github.com/jina-ai/jina/issues/763 we should need to add Chunk2DocRankerDriver explicitly in the yaml file?

YueLiu-jina:
Hi Joan, #148 I think these changes can avoid problems and make urbandict working normally. Although the broken query result is still un-fixed, this has no urgent now.

nan-wang:
Nope, we still need changes on the jina side. The more discussion is at
https://github.com/jina-ai/jina/issues/799

JoanFM:
Yes, but besides this the changes of the new commit are also needed?

urbandict-search/flow-query.yml
```diff
@@ -3,6 +3,10 @@ with:
   read_only: true
   port_expose: 45678
 pods:
+  sentencizer:
```
JoanFM, line 6: This is not needed, that is why we changed the `depth levels` at the rest of the Flow. The input of the search, will be a single sentence, so we do not want to `Sentencize`

urbandict-search/yaml/encode.yml
```diff
@@ -14,4 +14,4 @@ requests:
       - !EncodeDriver
         with:
           traverse_on: chunks
-          depth_range: [0, 0]
+          depth_range: [1, 1]
```
JoanFM, line 17: if Sentencizer is removed, this goes back to [0, 0]

urbandict-search/yaml/index-chunk.yml
```diff
@@ -35,10 +35,17 @@ requests:
           executor: vecidx
           top_k: 5
           traverse_on: chunks
-          depth_range: [0, 0]
```
JoanFM: if Sentencizer is removed, this goes back to [0, 0]

urbandict-search/yaml/index-chunk.yml
```diff
@@ -35,10 +35,17 @@ requests:
...
+          depth_range: [1, 1]
       - !KVSearchDriver
         with:
           executor: chunkidx
           key: id
-          traverse_on: matches
-          depth_range: [0, 1]
+          traverse_on: chunk_matches
+          depth_range: [1, 1]
```
JoanFM, line 44: if Sentencizer is removed, this goes back to [0, 1], or [0, 0] not sure

JoanFM:
Current logs of an example with 5 documents indexed.

[urbandict-issue-message-tracing.txt](https://github.com/jina-ai/examples/files/5056993/urbandict-issue-message-tracing.txt)

JoanFM:
This PR is ready for merge once in core this PR is merged:  https://github.com/jina-ai/jina/pull/851

JoanFM:
Ready to merge

Issue #116, opened 2020-07-27
examples/jina-0.4.1 migration


bwanglzu:
As was discussed, I'll start working on Jina examples migrations for Jina `0.4.1` release following the migration plan [here](https://github.com/jina-ai/jina/issues/702).

This issue is about to set up `0.4.1` migration for all the examples

|  Example  | Migrate-by  |  Peer-tested-by |  Status  | Pull Request |
|---|---|---|---|---|
|   face-db-search |  @bwanglzu  | @YueLiu-jina   |  Merged | #117  |
|  faiss-search | @bwanglzu   |   | Merged | #130   |
| flower-search  | @YueLiu-jina  |  |  Merged |  #119 |
|  helloworld-in-cs | @YueLiu-jina  |   | Merged  | #140  |
|  in-mock-app | @YueLiu-jina  |   |  Merged | #125  |
|  multilingual-search | @bwanglzu  |   | Merged  |  #127  |
| multires-lyrics-search  | @bwanglzu   |   | Merged  | #128   |
|  pokedex-with-bit | @bwanglzu   |  @YueLiu-jina | Merged  | #120  |
|  southpark-search  | @YueLiu-jina   | @bwanglzu   | Merged  | #119  |
| tumblr-gif-search  |  @bwanglzu  |   |  Merged | #129   |
| urbandict-search  | @YueLiu-jina  |   | Merged  |  #119 |
|  x-as-service | @bwanglzu   | @YueLiu-jina  | Merged  | #122  |
| zh/news-search  | @YueLiu-jina  |   | Problem  | #143  |
| zh/webqa-search  |  @YueLiu-jina |   | Merged  | #124  |

JoanFM:
We have to keep in mind to update the hub images to work with 0.4.1 as well

bwanglzu:
Hi @lusloher , since I don't have the jira access at the moment, if you wanna keep track the migration progress, please follow this channel.

JoanFM:
HelloWorld is merged now right?

YueLiu-jina:
Yes, news is the only one left

YueLiu-jina:
All examples except news-search have been migrated. News-search meets some problems that are non-trivial to solve with version 0.4.1. So maybe we will directly migrate it to 0.5 in another PR.

PR #140, opened 2020-08-05
Migrate helloworld in cs to 0.4.1

YueLiu-jina:
Part of #116 , ready for review


helloworld-in-cs/client.py
```diff
@@ -24,11 +26,12 @@ def hello_world(args):
...
     # download the data
-    download_data(targets)
+    download_data(targets, args.download_proxy)
+
 
     # run it!
-    py_client(port_expose=args.port_expose, host=args.host).index(
-        input_fn(targets['index']['filename']), batch_size=args.index_batch_size)
+    py_client(port_expose=args.port_expose, host=args.host, batch_size=args.index_batch_size).index(
+        input_numpy(targets['index']['data']), batch_size=args.index_batch_size)
```
JoanFM, line 34: targets['index''] does not have `data`

YueLiu-jina: Yes, I also get very confused at this. When I print `targets['index']['data']` I also get a key error. But `targets['index']['data']` is how it works, and `targets['index']['filename']`

YueLiu-jina: This change to 'data' is inspired by our Jina repo's helloworld code.

YueLiu-jina: targets['index']['data'] is add in the `download_data(targets, args.download_proxy)` function in the previous step

PR #143, opened 2020-08-06
Migrate news search to 0.4.1

YueLiu-jina:
part of #116 

Since the dataset is so large, processing the whole dataset may take too long. So a `valid` mode is added. One can use it by `python prepare_data.py valid`

This PR is far from finish, I will continue work on this. This news search meets the same problem as urbandict, even worse.

1. In `resp.search.docs`, some matches looks like following ,which contains no search result.

```
matches {
  id: 37
  weight: 1.0
  length: 32
  mime_type: "text/plain"
  score {
    value: 6.565842151641846
    op_name: "NumpyIndexer"
    ref_id: 1
  }
}
```

2. Some other matches have other problem:
```
matches {
  id: 26
  chunks {
    id: 1901
    weight: 1.0
    length: 3
    mime_type: "text/plain"
    text: "(www."
    level_depth: 1
    parent_id: 26
    location: 0
    location: 10
    embedding {
    }
  }
  chunks {
    id: 1902
    weight: 0.550000011920929
    length: 3
    mime_type: "text/plain"
    text: "ahnews."
    level_depth: 1
    parent_id: 26
    location: 10
    location: 17
    offset: 1
    embedding {
    }
  }
  chunks {
    id: 1903
    weight: 0.10000000149011612
    length: 3
    mime_type: "text/plain"
    text: "org)1025"
    level_depth: 1
    parent_id: 26
    location: 17
    location: 110
    offset: 2
    embedding {
    }
  }
  weight: 1.0
  length: 32
  mime_type: "text/plain"
  score {
    value: 6.933157444000244
    op_name: "NumpyIndexer"
    ref_id: 1
  }
}
```
This is unbelievable, because the whole dataset is in Chinese, while the `chunks.text` contains only alphabets and numbers.

A full version  of `resp.search.docs` is here
[resp.search.docs.txt](https://github.com/jina-ai/examples/files/5034208/resp.search.docs.txt)

### DIFF
--- a/urbandict-search/yaml/query-craft.yml
+++ b/urbandict-search/yaml/query-craft.yml
@@ -1,0 +1,5 @@
+!Sentencizer
+with:
+  min_sent_len: 3
+  max_sent_len: 128
+  punct_chars: '.,;!?:'
--- a/urbandict-search/yaml/query-craft.yml
+++ b/urbandict-search/yaml/query-craft.yml
@@ -1,5 +1,0 @@
-!Sentencizer
-with:
-  min_sent_len: 3
-  max_sent_len: 128
-  punct_chars: '.,;!?:'
--- a/zh/news-search/app.py
+++ b/zh/news-search/app.py
@@ -15,9 +15,11 @@ os.environ['TMP_WORKSPACE'] = workspace_path
 def print_topk(resp):
     print(f'以下是相似的新闻内容:')
     for d in resp.search.docs:
-        for tk in d.topk_results:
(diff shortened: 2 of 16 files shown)


bwanglzu:
I'm testing this locally, to see if I can reproduce

nan-wang:
hi, there. Is this because some temporal leftovers at /tmp/jina/news/?

YueLiu-jina:
@nan-wang I just deleted the whole thing and gave it a re-test. The same problem appeared. I think this is the same problem with urbandict

PR #144, opened 2020-08-06
remove docker image from index flow

theUnkownName:
No need to have the faiss docker image in index flow time


JoanFM:
Please @theUnkownName solve the conflicts before being able to merge this chanes

nan-wang:
This will require the installation of `faiss` which is overcomplicated for running an example.

JoanFM:
Notice that is just at index time when just NumpyIndexer is used

nan-wang:
I see. Then it is fine.

JoanFM:
It is now more `explicit` the fact that we can switch `indexer` at `query time`

PR #147, opened 2020-08-10
fix(southpark): change pooling strategy and fix bug

YueLiu-jina:
resolve #145 
resolve #146

PR #148, opened 2020-08-10
fix(urbandict): changing pooling strategy

YueLiu-jina:


urbandict-search/yaml/splitter.py
```diff
@@ -9,4 +9,4 @@
 class Splitter(BaseCrafter):
     def craft(self, text: str, *args, **kwargs) -> Dict:
         word, definitions = text.split('+-=')
-        return dict(text=definitions, meta_info=word.encode())
+        return dict(text=definitions, meta_info=text.encode())
```
JoanFM, line 12: why this? we are working on an update that will keep this, in the `meta_info` keeping the word

YueLiu-jina: I think before we fix the problem, we just use this way to make it work

YueLiu-jina: What's your consideration about this meta_info?

JoanFM: How is this used in the `app.py`?

YueLiu-jina: I'm thinking to do it the same way as southwark. The whole line word+-=definition will be kept in meta_info and be separate again in app.py so word and definition can be fetched

JoanFM: Word should be kept as `meta_info` because it is not part of the `embedding`. The separation will be done by finding the actual `word definitions` in the texts (that is what is actually embedded), and the `word` to define is just `meta_info` that is the word refering to `but not encoded`. I want to review southpark example to keep the `character` as `meta_info` so that `Filtering` can be applied to that field.

YueLiu-jina: What I'm thinking is that we still use `text=definition` to do the embedding. Because one text is corresponding to one meta_info, we can retrieve both word and definition from meta_info

YueLiu-jina: Like put definition both in text as embedding and in meta_info for retrieving

JoanFM: Yes, we could, but it is more `educational` to have in meta_info just the word. Why would u duplicate information?

JoanFM: besides, it is doing the split twice, is harder to keep track of the changes, if we change the split in one place we need to change in the other.

YueLiu-jina: Yes I agree, this is only aimed to avoid the problem and make it work first

JoanFM: If it is just as a workaround to have a working solution ASAP I am ok!

YueLiu-jina: made the proposed changes in #147 for southpark. Everything works well for southpark except the ugly `chunks[0]`

YueLiu-jina: seems urbandict is the only one that cuts text into different chunks

JoanFM: > made the proposed changes in #147 for southpark. Everything works well for southpark except the ugly `chunks[0]`

Yes, I think we have the same issue as in `urbandict`, will take a look afterwards

PR #150, opened 2020-08-11
Feat scann search

theUnkownName:


faiss-search/scann/scann.py
```diff
@@ -0,0 +1,45 @@
...
+__license__ = "Apache-2.0"
+
+from typing import Tuple
+
+import numpy as np
+
+from jina.executors.indexers.vector import BaseNumpyIndexer
+
+
+class ScannIndexer(BaseNumpyIndexer):
```
JoanFM, line 11: Cool, it would also be good to add unittests showcasing the usage

scann-search/app.py
```diff
@@ -0,0 +1,27 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
```
JoanFM, line 1: Remember to remove scann search folder

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,45 @@
...
+        self.num_leaves = num_leaves
+        self.num_leaves_to_search = num_leaves_to_search
+        self.training_sample_size = training_sample_size
+        self.scoring = scoring
+        self.anisotropic_quantization_threshold = anisotropic_quantization_threshold
+
+    def build_advanced_index(self, vecs: 'np.ndarray'):
+        """Load vectors into Scann indexers """
+        import scann
+        normalized_dataset = vecs / np.linalg.norm(vecs, axis=1)[:, np.newaxis]
```
JoanFM, line 42: they have this in their `example notebook`. But are we sure we need to `normalize`?

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,45 @@
...
+        _index = scann.ScannBuilder(normalized_dataset, 10, "dot_product").tree(self.num_leaves, self.num_leaves_to_search, self.train_filepath)
```
JoanFM, line 43: 10 and "dot_product" are parameters that we need to expose

JoanFM:
Take a look at `https://github.com/google-research/google-research/blob/master/scann/scann/scann_ops/py/scann_builder.py` to see what parameters we would like to expose for each method

faiss-search/scann/scann.py
JoanFM: and u need to have `score_ah` to atually build the index

theUnkownName: Ok, I was confused by this, I thought only the `.tree()` was needed to build the index, not the score, but yeah I'll change it.

faiss-search/scann/scann.py
```diff
@@ -18,28 +18,46 @@ class ScannIndexer(BaseNumpyIndexer):
...
+        self.dimensions_per_block = dimensions_per_block
 
     def build_advanced_index(self, vecs: 'np.ndarray'):
         """Load vectors into Scann indexers """
         import scann
-        normalized_dataset = vecs / np.linalg.norm(vecs, axis=1)[:, np.newaxis]
-        _index = scann.ScannBuilder(normalized_dataset, 10, "dot_product").tree(self.num_leaves, self.num_leaves_to_search, self.train_filepath)
+        _index = scann.ScannBuilder(vecs, self.training_iterations, self.distance_measure).\
+            tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
+            score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).reorder(100).create_pybind()
```
JoanFM, line 61: why is the reorder hardcoded?

faiss-search/scann/scann.py
```diff
@@ -18,28 +18,46 @@ class ScannIndexer(BaseNumpyIndexer):
...
+                 scoring: str = "score_ah",
+                 anisotropic_quantization_threshold: float = 0.2,
+                 dimensions_per_block: int = 2,
+                 *args, **kwargs):
         """
         Initialize an ScannIndexer
 
         :param num_leaves: The higher num_leaves, the higher-quality the partitioning will be. It should be roughly the square root of the number of datapoints.
         :param scoring: It can be score_brute_force and score_ah (asymmetric hashing)
+        :param training_iterations: Default is 10
```
JoanFM, line 36: Some explanation of the parameters?

faiss-search/scann/scann.py
```diff
@@ -18,28 +18,46 @@ class ScannIndexer(BaseNumpyIndexer):
         Scann package dependency is only required at the query time.
     """
 
-    def __init__(self, train_filepath: str = None, num_leaves: int = 2000, num_leaves_to_search: int = 100, training_sample_size: int = 250000,
-                 scoring: str = "score_ah", anisotropic_quantization_threshold: float = 0.2, *args, **kwargs):
+    def __init__(self, train_filepath: str = None,
```
JoanFM, line 21: Does it work with `train_filepath` to None?

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,85 @@
...
+        :param training_sample_size: The size of the training sample.
+            Default is 25k
+        :param scoring: It can be score_ah (asymmetric hashing) or score_bf for brute force.
+            For small datasets (less than 20k) brute force is recommended
+        :param anisotropic_quantization_threshold: See https://arxiv.org/abs/1908.10396 for
+            a description of this value
+        :param dimensions_per_block: Recommended for AH is 2
+        :param reordering_num_neighbors: Should be higher than the final number of neighbors
+            If this number is increased, the accuracy will increase but it will impact speed
+        :param args:
```
JoanFM, line 52: args and kwargs are not usually added as params

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,85 @@
...
+        self.distance_measure = distance_measure
+        self.num_leaves_to_search = num_leaves_to_search
+        self.training_sample_size = training_sample_size
+        self.scoring = scoring
+        self.anisotropic_quantization_threshold = anisotropic_quantization_threshold
+        self.dimensions_per_block = dimensions_per_block
+        self.reordering_num_neighbors = reordering_num_neighbors
+
+    def build_advanced_index(self, vecs: 'np.ndarray'):
+        """Load vectors into Scann indexers
```
JoanFM, line 68: They are using `Command Pattern` (https://sourcemaking.com/design_patterns/command). We should comment that is a lazy evaluation. That `score_ah` and `reorder` are creating just `configuration` to be passed to `create_py_bind` which is the one actually building the object.

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,85 @@
...
+                 training_sample_size: int = 250000,
+                 scoring: str = "score_ah",
+                 anisotropic_quantization_threshold: float = 0.2,
+                 dimensions_per_block: int = 2,
+                 reordering_num_neighbors: int = 100,
+                 *args, **kwargs):
+        """
+        Initialize an ScannIndexer
+
+        :param train_filepath: the training data file path.
```
JoanFM, line 35: remove `train_filepath`

faiss-search/scann/scann.py
```diff
@@ -66,20 +63,25 @@ def __init__(self, train_filepath: str = None,
...
         """
         import scann
 
         _index = scann.ScannBuilder(vecs, self.training_iterations, self.distance_measure).\
             tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
-            score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).reorder(self.reordering_num_neighbors).create_pybind()
-        return _index
+            score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).\
+            reorder(self.reordering_num_neighbors).create_pybind()
+        return self._index
```
JoanFM, line 84: there is no `self._index` and should not be assigned here

faiss-search/scann/scann.py
```diff
@@ -66,20 +63,25 @@ def __init__(self, train_filepath: str = None,
...
 
+    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
+        neighbors, distances = self._index.search_batched(keys)
```
JoanFM, line 87: since we assume that `reordering_num_neighbors`, we should add:

```python
if (self.reordering_num_neighbors < self.top_k):
    self.logger.warning('bla bla your reordering_num_neighbors should be larger or equal than ...')
```

And then, since `neighbors and distances` will have lentgh `self.reordering_num_neighbors` (to be tested), you need to select the `top_k`

```

File: faiss-search/scann/scann.py [hunk 13]
```diff
@@ -81,7 +83,9 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
             tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
```
JoanFM, line 83: sorry I did not express it correctly, you do retrun `self._index`, but you need to assign it, here `_index` is just a local variable

faiss-search/scann/scann.py
```diff
@@ -6,6 +6,7 @@
 import numpy as np
 
 from jina.executors.indexers.vector import BaseNumpyIndexer
+from jina.executors.frameworks import BaseFrameworkExecutor
```
JoanFM, line 9: Remove this import

faiss-search/scann/scann.py
```diff
@@ -49,6 +50,7 @@ def __init__(self,
             If this number is increased, the accuracy will increase but it will impact speed
         """
         super().__init__(*args, **kwargs)
+        self.searcher = None
```
JoanFM, line 53: This is normally so, but to avoid `jina` from serializing this object, this initialization should be avoided.

faiss-search/scann/scann.py
```diff
@@ -46,11 +45,10 @@ def __init__(self,
         :param anisotropic_quantization_threshold: See https://arxiv.org/abs/1908.10396 for
             a description of this value
         :param dimensions_per_block: Recommended for AH is 2
-        :param reordering_num_neighbors: Should be higher than the final number of neighbors
+        :param reordering_num_neighbors: Should be higher than the final number of neighborsjina
```
JoanFM, line 48: typo

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,94 @@
...
+        searcher = scann.ScannBuilder(vecs, self.training_iterations, self.distance_measure).\
+            tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
+            score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).\
+            reorder(self.reordering_num_neighbors).create_pybind()
+        return self.searcher
+
+    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
+        if self.reordering_num_neighbors < self.top_k:
+            self.logger.warning('The number of reordering_num_neighbors should be the same or higher than the number of neighbors')
+        neighbors, distances = self.searcher.search_batched(keys)
```
JoanFM, line 91: what value will `self.searcher` has here? Has it been assigned somewhere?

faiss-search/scann/scann.py
```diff
@@ -83,12 +83,12 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
             tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
             score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).\
             reorder(self.reordering_num_neighbors).create_pybind()
-        return self.searcher
+        return searcher
 
     def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
         if self.reordering_num_neighbors < self.top_k:
```
JoanFM, line 89: This condition should be reversed

faiss-search/scann/scann.py
```diff
@@ -0,0 +1,94 @@
...
+        searcher = scann.ScannBuilder(vecs, self.training_iterations, self.distance_measure).\
+            tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
+            score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).\
+            reorder(self.reordering_num_neighbors).create_pybind()
+        return searcher
+
+    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
+        if self.reordering_num_neighbors < self.top_k:
+            self.logger.warning('The number of reordering_num_neighbors should be the same or higher than the number of neighbors')
+        neighbors, distances = searcher.search_batched(keys)
```
JoanFM, line 91: We need to take care of getting only top-k? Where are we using this param?

faiss-search/scann/test_scann.py
```diff
@@ -0,0 +1,56 @@
...
+            save_abspath = a.save_abspath
+
+        with BaseIndexer.load(save_abspath) as b:
+            idx, dist = b.query(query, top_k=4)
+            global retr_idx
+            if retr_idx is None:
+                retr_idx = idx
+            else:
+                np.testing.assert_almost_equal(retr_idx, idx)
+            self.assertEqual(idx.shape, dist.shape)
```
JoanFM, line 49: Is it the shape we want to have?

faiss-search/scann/scann.py
theUnkownName: I thought `reordering_num_neighbors` be higher than k? They have this on their documentation:
"`reordering_num_neighbors` should be greater than k, the final number of neighbors."

JoanFM: yes sorry, my bad. Ignore this

faiss-search/scann/test_scann.py
JoanFM: Ignore this

faiss-search/scann/scann.py
theUnkownName: **We need to take care of getting only top-k?** I think so, the rescore takes only the best distances from scoring and uses only those to rescore
**Where are we using this param?** the neighbors one? this one is then used directly to compute the recall

JoanFM: what I mean is that reordering_num_neighbors distances will be returned by `search_batched`. But in the output of `query` we only want `top_k`

faiss-search/scann/scann.py
```diff
@@ -58,8 +62,9 @@ def __init__(self,
         self.anisotropic_quantization_threshold = anisotropic_quantization_threshold
         self.dimensions_per_block = dimensions_per_block
         self.reordering_num_neighbors = reordering_num_neighbors
+        self.final_num_neighbors = final_num_neighbors
 
-    def build_advanced_index(self, vecs: 'np.ndarray'):
+    def build_advanced_index(self,  dataset):
```
JoanFM, line 67: do not change this interface, since it is imposed by parent class

faiss-search/scann/scann.py
```diff
@@ -77,18 +82,16 @@ def build_advanced_index(self, vecs: 'np.ndarray'):
...
-        searcher = scann.ScannBuilder(vecs, self.training_iterations, self.distance_measure).\
+        searcher = scann.ScannBuilder(dataset, self.training_iterations, self.distance_measure).\
             tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size).\
             score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold).\
             reorder(self.reordering_num_neighbors).create_pybind()
         return searcher
 
-    def query(self, searcher, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
-        if self.reordering_num_neighbors < self.top_k:
+    def query(self, searcher, queries):
```
JoanFM, line 91: same here, do not change these signatures

faiss-search/scann_example/scann_example.py
```diff
@@ -0,0 +1,95 @@
...
+        index = scann.ScannBuilder(vecs, self.training_iterations, self.distance_measure). \
+            tree(self.num_leaves, self.num_leaves_to_search, self.training_sample_size). \
+            score_ah(self.dimensions_per_block, self.anisotropic_quantization_threshold). \
+            reorder(self.reordering_num_neighbors).create_pybind()
+        return index
+
+    def query(self, keys: 'np.ndarray', top_k: int, *args, **kwargs) -> Tuple['np.ndarray', 'np.ndarray']:
+        if self.reordering_num_neighbors < top_k:
+            self.logger.warning('The number of reordering_num_neighbors should be the same or higher than top_k')
+        print("*************** TEST query **********")
```
JoanFM, line 92: Please remove prints

faiss-search/scann_example/test_scann.py
```diff
@@ -0,0 +1,47 @@
...
+            print(idx, dist)
+            global retr_idx
+            if retr_idx is None:
+                retr_idx = idx
+            else:
+                np.testing.assert_almost_equal(retr_idx, idx)
+            self.assertEqual(idx.shape, dist.shape)
+            self.assertEqual(idx.shape, (10, 4))
+
+        #self.add_tmpfile(index_abspath, save_abspath)
```
JoanFM, line 43: please add this and add the functionality coming from `JinaTestCase` (Copy paste that functionality for the time being)

JoanFM: Plus add the test where reordering_num_neighbors is lower than `top_k`

faiss-search/scann_example/test_scann.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+if __name__ == '__main__':
```
JoanFM, line 46: Remove these part since we are currently using `pytest` (I know, we are mixing things but it should soon be gone)

JoanFM (changes requested):

nan-wang:
`Scann` should also be migrated to `hub`, right?

JoanFM:
yes, exactly

nan-wang:
related to https://github.com/jina-ai/jina-hub/pull/162

JoanFM:
@theUnknownName, after scann migration to hub we can still add Scann as an option to be used in this example and change the example name maybe

PR #152, opened 2020-08-12
examples: my first jina app

alexcg1:
A while ago I wrote a tutorial for brand new users to create a new app from scratch, using cookiecutter.

Since we're not pushing cookiecutter so much any more, I don't know how suitable this tutorial is, so I'm making a PR just in case it **is** useful.

Otherwise I can work out some other method to make it user-friendly without cookiecutter and re-submit


alexcg1:
I'll fix the image scaling - then can we merge? @hanxiao @nan-wang

hanxiao:
yes

alexcg1:
Done! (You may need to clear cache to see the change. I had to open page in another browser before I could see new version)

PR #157, opened 2020-08-17
chore: add new example to README

alexcg1:
Adding as first example in examples README, so new users know exactly where to start. Once this is merged I'll edit it into core repo README too

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@
 
 <tr>
 <td>
+<h4><a href="https://github.com/jina-ai/examples/tree/master/my-first-jina-app">My First Jina App</a></h4>
+Get started in your Jina journey by building a basic text search app and learning as you go
+</td>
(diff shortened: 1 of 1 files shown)


YueLiu-jina:
Hi Alex, a small suggestion here. I feel like this example goes pretty deep, so I think it's not proper to put it in the first of the list. Should we put it afterward, like to be the first of 🕊level, not 🐣 level? Or put it after the southpark one?

alexcg1:
This example goes deep, but it starts out holding the user's hand and guides them through every step. I feel the other examples are not so new user friendly

This was originally an attempt to rewrite the SP example, but I've since targeted it at new users, and included cookiecutter to make things easier

alexcg1:
Since cookiecutter is being fixed, this example doesn't work for now, hence closing PR until we get a working version

PR #158, opened 2020-08-18
merge facedb-readme into flower-readme

YueLiu-jina:
The motivation for this PR is written in #131 

Due to the high similarity between flower readme and facedb readme, this PR make the flower readme more detailed, add a section named 'similar examples' to replace the whole facedb readme.

If these changes are approved, I'll continue on merging other similar readme.

PR #163, opened 2020-08-21
refactor(face-db): have code consistency similar to other exam…

kaushikb11:
The motivation behind this PR is when a User is trying out multiple examples, code consistency & similar patterns could be helpful. Refactored some minor changes in the `face-db-search` example. README seems to be getting updated in #158.

### DIFF
--- a/face-db-search/app.py
+++ b/face-db-search/app.py
@@ -3,63 +3,43 @@ __license__ = "Apache-2.0"
 
 
 import os
-import sys
+import click
 
 from jina.flow import Flow
 
-num_docs = os.environ.get('MAX_DOCS', 500)
-image_src = '/tmp/jina/celeb/lfw/**/*.jpg'
-
 
 def config():
-    parallel = 2 if sys.argv[1] == 'index' else 1
-    shards = 8
-
     os.environ['TMP_WORKSPACE'] = '/tmp/jina/workspace'
     os.environ['COLOR_CHANNEL_AXIS'] = str(0)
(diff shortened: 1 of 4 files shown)


face-db-search/app.py
```diff
@@ -3,63 +3,43 @@
...
-def dryrun():
-    f = Flow.load_config('flow-query.yml')
-
-    with f:
-        pass
+@click.command()
+@click.option('--task', '-t')
+@click.option('--num_docs', '-n', default=500)
+def main(task, num_docs):
+    config()
```
JoanFM, line 23: would make more sense to have task as param of the config then?

face-db-search/app.py
```diff
@@ -3,63 +3,45 @@
...
 from jina.flow import Flow
 
-num_docs = os.environ.get('MAX_DOCS', 500)
-image_src = '/tmp/jina/celeb/lfw/**/*.jpg'
 
-
-def config():
-    parallel = 2 if sys.argv[1] == 'index' else 1
-    shards = 8
+def config(task):
```
nan-wang, line 11: `task` is not used.

nan-wang (changes requested):

face-db-search/app.py
kaushikb11: @nan-wang `parallel = 2 if task == 'index' else 1` There's this condition in the `config` function to set `PARALLEL` env variable.

nan-wang (approved):
LGTM👍

PR #166, opened 2020-08-24
refactor: remove ranker is not necessary

JoanFM:
For structures like this, it would be necessary (imagine here we would use NumpyIndexer at query time) to have the change in 
https://github.com/jina-ai/jina/pull/870


nan-wang (approved):
LGTM👍

PR #181, opened 2020-09-02
New example using BERT with jina

SirsikarAkshay:
Search for BCC news using the Jina Neural Search framework.


hanxiao:
this commits also binaries files

PR #197, opened 2020-09-14
fix: southpark runnable with next release

maximilianwerk:
changes:

- removed parallel and sharding to reduce noise in the example
- removed craft-split since there is no need for deeper document levels
- explanation from chunks removed, since there are now chunks anymore for this example
- new pictures. Please check if you like them or rather want the nice Mac-like-pictures for consistent look in all examples. If so, please any Mac user, make new pictures.

To be discussed:

- remove ranker completely, since the chunk-indexer already sorts results by distance?
- I did not update the README.restful.md. Do we want this to be polished this week?
- `Dive into the Pods` section is not yet polished. The reason is, that it explains heavily, how different driver are used. Do we really want this in this example, since it is not needed for the functionality, this example should have.
- This seems like the easiest example we have. Should we replace the my-first-jina-app with the southpark example and put the my-first-jina-app StarTrek example as a separate second example, since it really needs the nested document structure jina provides? 

A small comment on the style: I put single sentence in the `README.md` file on a single line whenever I edited them. This makes it easier to understand where changes where done in the git diff later on.


JoanFM:
I am all IN for removing `rankers` until the refactor hits in

alexcg1:
Re: Dive into Pods, I think that's way too much info for a beginner tutorial. I think it can be deleted (or at least collapsed)

Re: Replacing my first jina app, I'm all for it. As long as we keep same basic structure and use cookiecutter (as we discussed in our call)

alexcg1 (approved):
LGTM, though you should listen more to @JoanFM than me for technical side of things ;)

JoanFM (approved):

maximilianwerk:
Ah that merge was a bit to early. It'll only work, once 0.5.5 is released...

alexcg1:
Oh, damn. I'll look into unmerging

PR #199, opened 2020-09-14
Revert "fix: southpark runnable with next release"

alexcg1:
Reverts jina-ai/examples#197

I merged too early. This should (in theory) revert that change


maximilianwerk:
Since we want to stay at 0.5.4 for the Hackaton, let me fix the example and make it more verbose and runnable for 0.5.4.

JoanFM:
I think it will be 0.5.5 riight?

PR #209, opened 2020-09-17
Docs multires fix

alexcg1:
Clean up, adding missing commands

### DIFF
--- a/multires-lyrics-search/README.md
+++ b/multires-lyrics-search/README.md
@@ -19,7 +19,14 @@
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 
-## Download lyrics data
+## Use toy data
+
+We have included 1000 lyrics as toy data in [`toy-data`](toy-data).
+This data is ready to use with this example.
+
+## [Optional] Download full lyrics dataset
+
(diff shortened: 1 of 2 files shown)


maximilianwerk (approved):

multires-lyrics-search/README.md
```diff
@@ -19,7 +19,14 @@
...
+
+We have included 1000 lyrics as toy data in [`toy-data`](toy-data).
+This data is ready to use with this example.
+
+## [Optional] Download full lyrics dataset
+
+If you want to use the full dataset, please run:
 
 ```bash
 pip install kaggle
```
JoanFM, line 32: it is important to say that they need to have the kaggle API key. I put a notice like this in cxmodal search example linking to kaggle info on how-to download one

maximilianwerk: Fair point. I'll add it.

alexcg1: So, if you don't download the full data it just uses the toy data automatically? Don't have to set any environment variables or fiddle with app.py?

maximilianwerk: :+1: I thought this would be a bit more useful for users ;)

PR #220, opened 2020-09-18
docs(pokedex): explain running outside docker

alexcg1:
Explain how to run outside Docker as well as inside Docker


nan-wang (approved):
LGTM👍

PR #223, opened 2020-09-19
fix: correct python commands

alexcg1:
Fix Python commands, for example `python app.py index` instead of `python app.py`


JamesTang-jinaai (commented):
Looks fine to me, thanks Alex : )

PR #234, opened 2020-09-23
Fix cxmodal

JoanFM:


cross-modal-search/dataset.py
```diff
@@ -67,7 +68,7 @@ def collate_fn(data):
     return images, captions
 
 
-def get_data_loader(split, root, captions, batch_size=8, dataset_type='f30k', shuffle=True,
+def get_data_loader(split, root, captions, batch_size=8, dataset_type='f30k', shuffle=False,
```
JoanFM, line 71: this lets u focus on the same images and captions over and over making it easier to debug

PR #252, opened 2020-10-16
test: added object search test

bhavsarpratik:
Added test for object search which indexes 3 images and then queries the same to check if the query is the same as the first result. This uses `flow-query-original.yml`


object-search/requirements.txt
```diff
@@ -1 +1 @@
-jina[cv,http]==0.6.4
\ No newline at end of file
+jina[cv,http]==0.6.7
```
JoanFM, line 1: we will need to make usre that we install the latest version in the `ci.yml` right?

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,110 @@
...
+import shutil
+import subprocess
+
+from PIL import Image
+import pytest
+
+from jina.flow import Flow
+
+
+os.chdir(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..' ))
```
JoanFM, line 15: why do we need to change dir and not to pass `paths` relative to the file dir?

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,110 @@
...
+	f = set_flow()
+	with f:
+		for query in queries:
+			output = get_results(query)
+			matches = output['search']['docs'][0]['matches']
+			buffer = matches[0]['buffer'] #getting buffer of first match
+			assert query == buffer2png(buffer) #first match should be the query itself
+
+
+def test_cleanup():
```
JoanFM, line 109: we should use `tmpdir` fixture to avoid using this `rmtree`

JoanFM (changes requested):

object-search/tests/test_objectsearch.py
bhavsarpratik: The index and query flow reference pod yml paths and so we need to change the path to root of example.

JoanFM: is it that so? are we sure? it feels like a smell (I am not sure)

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,110 @@
...
+	os.environ['SHARDS'] = str(shards)
+	os.environ['WORKDIR'] = './workspace'
+	os.makedirs(os.environ['WORKDIR'], exist_ok=True)
+
+
+def prepare_data():
+    pass
+
+
+def setup_env():
```
JoanFM, line 39: is there a reason to do it here and not to do it in the github action?

southpark-search/tests/test_southparksearch.py
```diff
@@ -22,19 +21,19 @@
 query_flow_file_path = "flow-query.yml"
 
 
-def config():
-	os.environ["JINA_DATA_FILE"] = os.environ.get("JINA_DATA_FILE", "tests/data-index.csv")
-	os.environ["JINA_WORKSPACE"] = os.environ.get("JINA_WORKSPACE", "workspace")
-	os.environ["JINA_PORT"] = os.environ.get("JINA_PORT", str(45678))
+def config(tmpdir):
+    os.environ["JINA_DATA_FILE"] = "tests/data-index.csv"
```
JoanFM, line 25: single quote if possible to keep style

southpark-search/tests/test_southparksearch.py
```diff
@@ -22,19 +21,19 @@
...
 
 
 def prepare_data():
     pass
 
 
 def setup_env():
-	subprocess.check_call([sys.executable, "-m", "pip", "install", "-r", "requirements.txt"])
-	subprocess.check_call([sys.executable, "-m", "pip", "install", "requests", "jina[http]"])
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "-r", "requirements.txt"])
```
JoanFM, line 35: if possible to add single quote all around

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,105 @@
...
+
+def index_documents():
+    f = Flow().load_config(index_flow_file_path)
+
+    with f:
+        f.index_files(os.environ["JINA_DATA"], batch_size=1, read_mode='rb', size=num_docs)
+
+
+def image_to_byte_array(image, format):
+        img_byte_arr = io.BytesIO()
```
JoanFM, line 50: something weird here with the `indentation`

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def get_results(query, top_k=top_k):
+    return call('post',
+                'http://0.0.0.0:45678/api/search', 
+                payload={"top_k": top_k, "mode": "search",  "data": [query]}
+                )
+
+
+def set_flow():
```
JoanFM, line 75: `get_flow`

southpark-search/tests/test_southparksearch.py
```diff
@@ -49,46 +49,41 @@ def index_documents():
...
-def test_cleanup():
-	shutil.rmtree(os.environ['JINA_WORKSPACE'])
+    return [('hey dude', ["Don't say anything\n", "Check that: I'll watch that game.\n", "Because you're a fucking fatass\n"]),
+            ('sister', ['Sorry\n', 'Lame.\n', 'Christ\n']),
+            ("Ill watch that game", ["Check that: I'll watch that game.\n", 'Quit it\n', 'Sorry\n'])]
+
+
+def test_query(queries, tmpdir):
+    config(tmpdir)
+    prepare_data()
```
JoanFM, line 78: I do not understand why do we do `prepare_data` if we are passing it?

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,105 @@
...
+
+def prepare_data():
+    pass
+
+
+def setup_env():
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "-r", "requirements.txt"])
+    subprocess.check_call([sys.executable, "-m", "pip", "install", "requests", "jina[http]"])
+
+setup_env()
```
JoanFM, line 39: This should be called inside the `test_function` or as a `fixture` (see how in some tests in the core we are setting env variable and deleting with a `generator` trick).

Like this, we would avoid any file importing it to run the `setup_env` when is not needed.

JoanFM (changes requested):

object-search/tests/test_objectsearch.py
JoanFM: I think this is important because I think `pytest` may import all the files in there and then filter the functions starting by `test_`, thus you may conflict pip installs depending on the order of importing and not depending on which example is being run

JoanFM:
Locally `test_southparksearch` is giving me the following resutls:

```
E               assert ['Quite an amazing creature.\n', "'Scuse me\n", "Nothing's changed.\n"] == ["Don't say anything\n",\n "Check that: I'll watch that game.\n",\n "Because you're a fucking fatass\n"]
E                 At index 0 diff: 'Quite an amazing creature.\n' != "Don't say anything\n"
E                 Full diff:
E                   [
E                 -  "Don't say anything\n",
E                 -  "Check that: I'll watch that game.\n",
E                 -  "Because you're a fucking fatass\n",
E                 +  'Quite an amazing creature.\n',
E                 +  "'Scuse me\n",
E                 +  "Nothing's changed.\n",
E                   ]

```

southpark-search/tests/test_southparksearch.py
bhavsarpratik: I wanted to keep the whole code as a template to be used for other examples. Initially, I was preparing the actual original data in Southpark and then use it but now that we are making custom small data for all examples, I will remove these.

JoanFM: Makes sense, good idea, just maybe we should add a comment there to guide potential future devs?

object-search/tests/test_objectsearch.py
bhavsarpratik: Earlier I kept `setup_env` as test_setup_env() but it felt weird because it wasn't a test. Then I moved it inside the `test_query` but it threw PIL import error due to import of fixture `queries` in the arguments of `test_query`. I like the fixture idea and its working if I make `setup_env` as fixture and pass it as first argument to `test_query`. 

I also moved chdir to this fixture and it has solved the object path problem too :-)

JoanFM: Setup env as fixture sounds good

bhavsarpratik:
@JoanFM I created a clean conda env, removed all transformer model cache and the tests ran successfully. Let's see if we can catch any error in the CI

southpark-search/tests/test_southparksearch.py
```diff
@@ -1,44 +1,36 @@
...
+    os.environ['JINA_WORKSPACE'] = str(tmpdir)
+    os.environ['JINA_PORT'] = str(45678)
 
 
+@pytest.fixture
 def setup_env():
-    subprocess.check_call([sys.executable, "-m", "pip", "install", "-r", "requirements.txt"])
-    subprocess.check_call([sys.executable, "-m", "pip", "install", "requests", "jina[http]"])
-
+    os.chdir(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..'))
```
JoanFM, line 25: should it be changed back afterwards'

bhavsarpratik: All tests will require to use the pods and will require cwd as the example's root and will have to run this chdir to be there. So I don't think we need to change it back to examples root although no harm in doing it.

southpark-search/tests/test_southparksearch.py
bhavsarpratik: I have removed it for now as we use manually made small datasets. So we don't need prepare_data anymore

JoanFM:
Should we run this offline and store the expected results and check against expected results? Right now the logic is fragile because since the semantic search is done at chunk level, there is no reason to be matched the same image as the `query`. So it is at least missleading

bhavsarpratik:
That is true. The problem we are facing is the local results are different than cloud results. The test is working on local and if we saved the png data and checked again, it will fail on workflow.

JoanFM:
> That is true. The problem we are facing is the local results are different than cloud results. The test is working on local and if we saved the png data and checked again, it will fail on workflow.

Yes, but it is also a problem that if I read the example I get a wrong impression of what the example does

object-search/tests/test_objectsearch.py
```diff
@@ -92,4 +93,5 @@ def test_query(setup_env, tmpdir, queries):
             output = get_results(query)
             matches = output['search']['docs'][0]['matches']
             buffer = matches[0]['buffer'] #getting buffer of first match
-            assert query == buffer2png(buffer) #first match should be the query itself
+            # assert query == buffer2png(buffer) #first match should be the query itself
+            buffer2png(buffer) #checking buffer conversion
```
JoanFM, line 97: this assert gives no insurance of what is happening, can we pass some `metadata` to the image at `index` time so that we can check that is the one we extract?

JoanFM (changes requested):

object-search/tests/test_objectsearch.py
bhavsarpratik: I think this assertion makes sure that the whole flow is geting executed(which checks Jina core+hub) but we cannot ensure the quality of results. Checking for metadata will be as good as the original query == buffer assertion because if we knew what metadata to check we can do the same by buffer. Let us merge this PR for now as we know the object search code is working and there is something causing discrepancy between local and workflow results.

JoanFM: Well, it would be different, if we use a `metadata` or `flag` that we know what the output should be makes sense, but here we just check that there is a `buffer` which can be converted to a `png`, but it can be very wronf

JoanFM: We can merge but we have to know how to change this or what we plan to do in another PR

bhavsarpratik: We would decide the meta based on the results we got on the local right? But the results coming in workflow are different than local and so the meta checking will fail.

JoanFM: I am lost, so we need to have a deeper investigation since we may be seeing a very deep bug

bhavsarpratik: Possible! Can you run this on your local in free time and confirm. I use Mac and having it confirmed on your Linux will confirm if it is a problem from workflow.

JoanFM: It works for me in local, but what does not work for me is the `southpark` test

JoanFM: but anyway, the test that assumes that `query` is the `output` looks like a wrong assumption and can be missleading on how the example works when looking at the test.

bhavsarpratik: That's weird! Different tests failing at different places.

JoanFM: Do a tests for both cases (southpark and object) and run the tests while adding `LogInfoDriver` at specific parts of the Flow and see where it is divergint

southpark-search/tests/test_southparksearch.py
maximilianwerk: I don't like this. Using `os.chdir` always is a code/setup smell for me. Why not prefix all needed files with `southpark-search`, if the tests are always run from examples main directory? With this, you don't need to use such a magic.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install jina[cicd,test] --no-cache-dir
+      - name: Test examples with pytest
+        run: |
+          jina check
+          export JINA_LOG_VERBOSITY="ERROR"
+          pytest -s -v */tests
```
maximilianwerk, line 30: southpark and object-search should not be tested in the same environment. We have to setup a fresh one per example. Otherwise the dependencies can collide.

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,97 @@
...
+import sys
+import shutil
+import subprocess
+
+import pytest
+
+from jina.flow import Flow
+
+
+num_docs = 3
```
maximilianwerk, line 14: Please uppercase all constants.

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,97 @@
...
+top_k = 3
+index_flow_file_path = 'flow-index.yml'
+query_flow_file_path = 'flow-query-original.yml'
+
+
+def config(tmpdir):
+    os.environ['JINA_DATA'] = 'tests/test-data/*.jpg'
+    os.environ['JINA_PORT'] = str(45678)
+    os.environ['PARALLEL'] = str(1)
```
maximilianwerk, line 23: Please prefix all Jina related environment variables with `JINA_`.

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,97 @@
...
+    os.environ['SHARDS'] = str(1)
+    os.environ['WORKDIR'] = str(tmpdir)
+
+
+@pytest.fixture
+def setup_env():
+    os.chdir(os.path.join(os.path.dirname(os.path.abspath(__file__)), '..' ))
```
maximilianwerk, line 30: Using `os.chdir` creates bugs for the future which are sometimes unforeseeable. I prefer assuming the tests are always run from the root directory `examples`. As far as I can see, this `od.chdir` mainly happens in order to have correct file paths. Ideally, just prefix all file paths with `object-search` and that should do the trick.

.github/workflows/ci.yml
JoanFM: so one action per test? or is there a way to adapt it with a unique action?

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,97 @@
...
+def read_and_convert2png(file_path):
+    from PIL import Image
+    im = Image.open(file_path)
+    png_bytes = image_to_byte_array(im, format='PNG')
+    return 'data:image/png;base64,' + base64.b64encode(png_bytes).decode()
+
+
+def call(method, url, payload=None, headers={'Content-Type': 'application/json'}):
+    import requests
+    return getattr(getattr(requests, method)(url, data=json.dumps(payload), headers=headers), 'json')()
```
maximilianwerk, line 58: I prefer not using reflection here but directly call the methods. Where is the advantage of using `method` as a parameter? Just do
```
requests.post(
    url, 
    data=json.dumps(payload), 
    headers={'Content-Type': 'application/json'}
).json()
```
Since the headers are even never used somewhere else, you can also hard code it here, until you need it in a calling function (which most probably it never will).

maximilianwerk (changes requested):

.github/workflows/ci.yml
maximilianwerk: I guess you can parametrize this in the matrix above with the directory names. Not sure how, but that should be possible.

JoanFM: Oh good idea, yes ! It is the way we test for `py3.7` , `py3.8`

bhavsarpratik: Let me check on this

object-search/tests/test_objectsearch.py
bhavsarpratik: I mainly had to do this because the `flow` refers to `pods` from the root of example.

.github/workflows/ci.yml
JoanFM: If we do this, then we have `pip install there` and avoid doing it from inside the `test` itself

.github/workflows/ci.yml
```diff
@@ -22,12 +23,14 @@ jobs:
...
-          pip install jina[cicd,test] --no-cache-dir
+          pip install jina[cicd,test,http] requests --no-cache-dir
+          pip install -r ${{ matrix.examples }}/requirements.txt --no-cache-dir
       - name: Test examples with pytest
         run: |
           jina check
           export JINA_LOG_VERBOSITY="ERROR"
-          pytest -s -v */tests
+          cd ${{ matrix.examples }}
+          pytest -s -v .
```
JoanFM, line 33: Better pass the path to the arguments

.github/workflows/ci.yml
```diff
@@ -0,0 +1,37 @@
...
+        with:
+          submodules: true
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install jina[cicd,test,http] requests --no-cache-dir
```
JoanFM, line 26: Do we have specific packages for cicd?

.github/workflows/ci.yml
bhavsarpratik: Do you mean rename `examples` in matrix to `path`?

.github/workflows/ci.yml
bhavsarpratik: `cicd` in used in jina's ci and it has many packages in `extra-requirements`

.github/workflows/ci.yml
JoanFM: No, I mean call `pytest -v -s ${{matrix.examples}}` instead of doing cd $...

.github/workflows/ci.yml
JoanFM: Good to know !!

bhavsarpratik: But I removed it and lets see if it works. I think we don't need it

bhavsarpratik:
@JoanFM Object search test passed the first time and now it is failing again

.github/workflows/ci.yml
bhavsarpratik: We need the `cd` for it to access the pods. I removed all `chdir` from the code and so need to do cd here

bhavsarpratik:
Now it passed all!

JoanFM:
This PR is not ready because the test does not reflect how the example should work, we should pass a chunk inage and get its original. Right now it works by chance.

bhavsarpratik:
For the same reason, I selected 3 images for the test which have big and different type of objects in it. So I find the current test trustable for checking Jina changes. The test varies a bit from the intended use but is a good hack because the query image is almost object size.

JoanFM:
> For the same reason, I selected 3 images for the test which have big and different type of objects in it. So I find the current test trustable for checking Jina changes. The test varies a bit from the intended use but is a good hack because the query image is almost object size.

I know it is trustworthy but it is a hack and it does not serve as documentation or shows how it works. Is it possible to pass these images once through the ObjectDetectionSegmenter and store them and use them as queries?

bhavsarpratik:
I can run the `query-object` flow and get the images via jinabox manually saving them. Then run the same flow and assert them.

JoanFM:
> I can run the `query-object` flow and get the images via jinabox manually saving them. Then run the same flow and assert them.

Yes, do it manually and then assert it in the test, good idea

bhavsarpratik:
@JoanFM Now the test does exactly what we discussed. I reran the test twice to make object search pass. There is some problem with it. Never fails on my local. Check, approve and merge this if its ready. I might have power cut in the morning-afternoon and so won't be able to do it.

JoanFM:
> @JoanFM Now the test does exactly what we discussed. I reran the test twice to make object search pass. There is some problem with it. Never fails on my local. Check, approve and merge this if its ready. I might have power cut in the morning-afternoon and so won't be able to do it.

What is the problem? Does it pass the tests?

bhavsarpratik:
This closes jina-ai/jina#1142

bhavsarpratik:
The object search fails sometimes on workflow and not on local

object-search/tests/test_objectsearch.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_query(tmpdir, image_paths):
+    config(tmpdir)
+    index_documents()
+    f = get_flow()
+    with f:
+        for query_image_path, object_image_path in image_paths:
+            query_image = read_and_convert2png(query_image_path)
+            object_image = read_and_convert2png(object_image_path)
+            output = get_results(query_image)
```
JoanFM, line 79: The query should be made with the object image right?

JoanFM:
> The object search fails sometimes on workflow and not on local

The query should be made with the object images.

object-search/tests/test_objectsearch.py
bhavsarpratik: Oops, my bad!

JoanFM (approved):
LGTM :+1:

JoanFM:
@bhavsarpratik it looks good to me, as per the assertion error itself, could it be that u need to extract the real path from the `uri`? It seems that the `uri` contains a little more information?

bhavsarpratik:
The test actually works on my local. In the workflow, the URI and input png data are very different. Only the first few characters are matching

JoanFM:
> The test actually works on my local. In the workflow, the URI and input png data are very different. Only the first few characters are matching

any clue why?

.github/workflows/ci.yml
maximilianwerk: Yes, I like this solution much more, than how it was before. Great job! :+1:

PR #263, opened 2020-11-03
Example southpark distributed

maximilianwerk:
This adds the ability to test a distributed southpark version. It would be nice if someone can test it via:
```
cd southpark-test
./tests/distributed/test_integration.sh
```
You may need to install docker-compose beforehand (see https://docs.docker.com/compose/install/ )


southpark-search/tests/distributed/test_integration.sh
```diff
@@ -0,0 +1,33 @@
...
+    -F "uses_files=@pods/index.yml" \
+    -F "pymodules_files=@pods/text_loader.py" \
+    -F "yamlspec=@tests/distributed/flow-query.yml"\
+    | jq -r .flow_id)
+
+echo "Successfully started the flow: ${FLOW_ID}"
+
+curl --request POST -d '{"top_k": 10, "mode": "search",  "data": ["text:hey, dude"]}' -H 'Content-Type: application/json' '0.0.0.0:45678/api/search'
+
+curl --request GET "http://0.0.0.0:8000/v1/flow/${FLOW_ID}" -H "accept: application/json"
```
JoanFM, line 29: what does this do?

theUnkownName:
Do I need some special access? it gives me this error: `ERROR: pull access denied for south, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

southpark-search/tests/distributed/.env
```diff
@@ -0,0 +1,3 @@
+JINA_ENCODER_HOST=encoder
```
JoanFM, line 1: `encoder` is a valid `host`?

JoanFM (commented):
What is the structure `approximated` in this example? It has the `flow` in one "server" and all the pods in another one?

southpark-search/tests/distributed/.env
maximilianwerk: in docker-compose it is.

> Do I need some special access? it gives me this error: `ERROR: pull access denied for south, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

fixed.

maximilianwerk:
> Do I need some special access? it gives me this error: `ERROR: pull access denied for south, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

Thanks for bringing this up. Fixed.

southpark-search/tests/distributed/test_integration.sh
maximilianwerk: Just a status check. Since we can...

maximilianwerk:
> What is the structure `approximated` in this example? It has the `flow` in one "server" and all the pods in another one?

It is encoder and indexer pods on the remote machine. The reason is, that I had issues putting one of them on the local machine. I'll try to achieve this in a further step.

theUnkownName (approved):
It looks good to me

JoanFM:
Would it be hard to go the extra step of adding a third "server" where the `index` could live?

maximilianwerk:
> Would it be hard to go the extra step of adding a third "server" where the `index` could live?

Nope, super easy, I'll do.

maximilianwerk:
Reasoning for fixing versions: That should be a blueprint for now for the terraform usage and in the jina CI. Thus, I want to avoid having non-running parts, when some updates happening to jina/jinad for now. In the future, the current jina/jinad versions should be used.

southpark-search/tests/distributed/test_integration.sh
```diff
@@ -0,0 +1,34 @@
+set -e
+
+if [ "${PWD##*/}" != "southpark-search" ]
+  then
+    echo "test_integration.sh should only be run from the southpark-search base directory"
+    exit 1
+fi
+
+export $(grep -v '^#' tests/distributed/.env | xargs -d '\n')
```
JoanFM, line 9: is this the line passing the `env variables`?

maximilianwerk: no, It sets them locally for this shell context. added a comment.

PR #264, opened 2020-11-03
test: multires lyrics search

cristianmtr:
Add integration test for multires lyrics search example, with assertion on chunk-level


JoanFM (commented):
Looks really good! 

I think it is important to do assertions at `chunks` matches fore this example

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -11,42 +11,44 @@
...
 def config(tmpdir):
-    parallel = 2 if sys.argv[1] == 'index' else 1
+    parallel = 2 if sys.argv[1] == "index" else 1
 
-    os.environ.setdefault('JINA_MAX_DOCS', '100')
-    os.environ.setdefault('JINA_PARALLEL', str(parallel))
-    os.environ.setdefault('JINA_SHARDS', str(4))
-    os.environ.setdefault('JINA_WORKSPACE', './workspace')
-    os.makedirs(os.environ['JINA_WORKSPACE'], exist_ok=True)
+    os.environ.setdefault("JINA_MAX_DOCS", "100")
```
JoanFM, line 21: better use `single quote` as is `jina style standard`

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -58,9 +60,35 @@ def get_flow():
 
 @pytest.fixture
 def queries():
-    return [("Trudging slowly\n", ["Don't say anything\n", "Check that: I'll watch that game.\n", "I'm trying to talk\n"]),
-            ('I could feel at the time\n', ['Sorry\n', 'Hey\n', 'Allб\n']),
-            ("I promise.\n", ["Check that: I'll watch that game.\n", "I just think it's a fabulous app\n", 'Could you get that\n'])]
+    return [
```
JoanFM, line 63: what is the `structure` of these `queries`? Aren't queries supposed to be provided as simple `strings`?

JoanFM: ah I see, can we rename `queries_with_results`?

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: Ah ok. This was `black`'s doing.

JoanFM: keep it, we will see

JoanFM (changes requested):
It looks very good, now I would try to observe the `matches` of the `chunks` and see if we can make sense and do some tests on them also.

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+    if headers is None:
+        headers = {"Content-Type": "application/json; charset=utf-8"}
+    import requests
+
+    return requests.post(url, data=json.dumps(payload), headers=headers).json()
+
+
+def get_results(query, top_k=TOP_K):
+    return call_api(
+        "http://0.0.0.0:45678/api/search", payload={"top_k": top_k, "data": [query]}
```
nan-wang, line 67: Let's convert `45678` into a variable. 
```suggestion
        "http://0.0.0.0:45678/api/search", payload={"top_k": top_k, "data": [query]}
```

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+import json
+import os
+import sys
+import subprocess
+
+from jina.flow import Flow
+import pytest
+from jina.proto import jina_pb2
+
+NUM_DOCS = 100
```
nan-wang, line 15:
```suggestion
```

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+
+    os.makedirs(os.environ["JINA_WORKSPACE"], exist_ok=True)
+    return
+
+
+def index_documents():
+    f = Flow().load_config(INDEX_FLOW_FILE_PATH)
+
+    with f:
+        f.index(input_fn, batch_size=8)
```
nan-wang, line 54: Keep it simple
```suggestion
        f.index(input_fn)
```

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def call_api(url, payload=None, headers=None):
+    if headers is None:
+        headers = {"Content-Type": "application/json; charset=utf-8"}
+    import requests
```
nan-wang, line 60: this is not included in `requirements.txt` and make sure it is available for cicd

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+
+def test_query(tmpdir, queries_and_expected_replies):
+    config(tmpdir)
+    index_documents()
+    f = get_flow()
+    with f:
+        for query, exp_result in queries_and_expected_replies:
+            output = get_results(query)
+            matches = output["search"]["docs"][0]["matches"]
+            print(f"matches = {matches}")
```
nan-wang, line 217:
```suggestion
```

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+                "you'll remember. When May is rushing over you. With desire to be part of the "
+                "miracles. You see in every hour. You'll know it's true. That you are blessed "
+                "and lucky. It's true that you are touched. By something that will grow and "
+                "bloom in you. These are days. These are the days you might fill. With "
+                "laughter until you break. These days you might feel. A shaft of light. Make "
+                "its way across your face. And when you do. You'll know how it was meant to "
+                "be. See the signs and know their meaning. You'll know how it was meant to "
+                "be. Hear the signs and know they're speaking. To you, to you",
+            ],
+        ),
```
nan-wang, line 105: Maybe we can figure out a better way to include the expected results.

multires-lyrics-search/tests/test_multireslyrics.py
maximilianwerk: Configure black with `black -S`, which will skip the string normalization.

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,224 @@
...
+                d.text = row[3]
+                yield d
+
+
+def config(tmpdir):
+    parallel = 2 if sys.argv[1] == "index" else 1
+
+    os.environ.setdefault("JINA_MAX_DOCS", "100")
+    os.environ.setdefault("JINA_PARALLEL", str(parallel))
+    os.environ.setdefault("JINA_SHARDS", str(1))
```
maximilianwerk, line 41: While for initial testing, this is nice, it would be great to have a (separated?) test with `JINA_SHARDS=2`. The reason: In the past, I saw, that results were sometimes flaky when sharding was used. If this is the case in this test, we would directly have spotted a bug and something to work on.

multires-lyrics-search/tests/test_multireslyrics.py
maximilianwerk: One possibility would be: Just test the first 5 of each list. If they are the same, most probably everything is fine.

JoanFM: I think we should index smaller lyrics. But also the tests shoould be run with `multi-chunk` queries to test the full power of the example and to have a good tests that heavily uses the `document recursivity`.

JoanFM: at the end we should "mimic" the color behavior observed in the frontend

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: I replaced all the double with single-quotes, apart from the query - expected results (where double quotes are required because there are singles within the text)

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: I will dig in and implement the assertion on the chunk level

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: Done

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: Good point

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: It was working without it being explicitly added, but I added it now. Might have been installed by another lib

multires-lyrics-search/tests/test_multireslyrics.py
cristianmtr: Leftover from logging. Cleaned up now

multires-lyrics-search/tests/test_multireslyrics.py
```diff
@@ -0,0 +1,172 @@
...
+
+def get_flow():
+    f = Flow().load_config(QUERY_FLOW_FILE_PATH)
+    f.use_rest_gateway()
+    return f
+
+
+@pytest.fixture
+def queries_and_expected_replies():
+    return [
```
JoanFM, line 79: Looks very good, I would just test then the direct resulting `match` of the `query`. Like this the test would look complete for me.

cristianmtr: You mean re-adding the assertion on the `'match'`-level, apart from the chunk-level, as it was before?

JoanFM: yes, the `multi-level` assertion. It may look a little redundant but is a test so better to make sure everything makes sense

cristianmtr: Done

maximilianwerk (approved):
LGTM Well done.

JoanFM (commented):
LGTM! I think we just need now to add it to the `CI workflow`!

cristianmtr:
> LGTM! I think we just need now to add it to the `CI workflow`!

Added it. Should be running now

JoanFM (approved):
Great work!

maximilianwerk:
> > LGTM! I think we just need now to add it to the `CI workflow`!
> 
> Added it. Should be running now
> 
> EDIT: And it passed partying_face

That is great. Thanks for the effort! We might want to update to jina 0.7.4 in a next iteration, but merge away for now.

PR #267, opened 2020-11-04
feat: adjust multimodal encoder

bwanglzu:
1. Change TIRGMultimodalEncoder to make it work.
2. Add instance level tests for TIRGEncoder and TIRGMultiModalEncoer.

I'll add batch level test soon.


multimodal-search-tirg/TirgMultimodalEncoder /__init__.py
```diff
@@ -52,25 +53,24 @@ def _get_features(self, data):
...
         text_features = self.model.extract_text_feature(textual_data)
-        
-        return self.model.compose_img_text_features(img_features, text_feature)
+
+        return self.model.compose_img_text_features(img_features, text_features)
 
     @batching
     @as_ndarray
-    def encode(self, *data: 'np.ndarray', args, **kwargs) -> 'np.ndarray':
+    def encode(self, data: list, *args, **kwargs) -> 'np.ndarray':
```
JoanFM, line 70: this signature should be:

```
def encode(self, *data: 'np.ndarray', **kwargs) -> 'np.ndarray':
```

JoanFM (changes requested):

JoanFM:
Why isn't here the `notebook` so that @bhavsarpratik can work on it? or the proper tests in the `executors`?

multimodal-search-tirg/TirgMultimodalEncoder /__init__.py
JoanFM: the point is we added also `*args` which can't be done

bwanglzu: done

bhavsarpratik (approved):

PR #278, opened 2020-11-10
feat: multimodal search tirg bo 2

bwanglzu:
@bhavsarpratik @JoanFM the multimodal encoder test passed locally and in docker, while image encoder passed locally , failed in docker. I'll put what I have now in this PR.


bwanglzu:
Close because we switch to css dataset in #287

Issue #276, opened 2020-11-09
multires lyrics crashes erratically in CI


cristianmtr:
I have noticed that the integration test for the multires lyrics example seems to be crashing randomly in CI.

This happens even though the PR in question had nothing to do with it, and even though locally I wasn't able to reproduce it. I have only seen this error once before locally, and it was on a previous version of Jina (`0.6.4`). It is now running on `0.7.4`. It is also using sharding (shards = 2). 

Any ideas on what might be causing this?

Stack trace looks like:

```
=================================== FAILURES ===================================
__________________________________ test_query __________________________________

tmpdir = local('/tmp/pytest-of-runner/pytest-0/test_query0')
queries_and_expected_replies = [['Take me now, baby, here as I am. Hold me close, and try and understand. Desire is hunger is', {'chunk-level': [{'ch...e). Everyday is like sunday. Win yourself a cheap tray. Share some grease tea with me. Everyday is silent and grey']}]]

    def test_query(tmpdir, queries_and_expected_replies):
        config(tmpdir)
        index_documents()
        f = get_flow()
        with f:
            for query, exp_result in queries_and_expected_replies:
                output = get_results(query)
    
                # chunk-level comparison
                chunks = output['search']['docs'][0]['chunks']
                query_chunk_results = []
                for chunk in chunks:
                    chunk_result = {'chunk': chunk['text'], 'chunk_matches': []}
                    chunk_matches = chunk['matches']
                    # make sure to sort in asc. order, by score
>                   chunk_matches = sorted(chunk_matches, key=lambda x: x['score']['value'], reverse=False)

tests/test_multireslyrics.py:99: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = {'adjacency': 1, 'embedding': {'dense': {'buffer': 'JRi6Pqs7Gj4gTd48mqfpvMojgr6UDxC+YQmQPsLyhrwJIFg9c7eavysGy77iG3U9rQ...y79LqjPn6hMj4H0gU+z7MMvnZrrT5u3Fe9', 'dtype': '<f4', 'shape': [768]}}, 'granularity': 1, 'id': '21b282d50a2593c2', ...}

>   chunk_matches = sorted(chunk_matches, key=lambda x: x['score']['value'], reverse=False)
E   KeyError: 'value'

[...]

PR #279, opened 2020-11-11
test: print failing chunk

maximilianwerk:
In order to investigate #276 we want to print whichever chunks are failing.


cristianmtr:
Initially there were two successful runs. It finally failed in the third run 

https://github.com/jina-ai/examples/runs/1385310435?check_suite_focus=true#step:5:667

PR #280, opened 2020-11-11
feat: add warning for existing workspace

theUnkownName:


southpark-search/app.py
```diff
@@ -81,7 +81,12 @@ def main(task, num_docs, top_k):
     workspace = os.environ["JINA_WORKSPACE"]
     if task == "index":
         if os.path.exists(workspace):
-            print(f"The directory {workspace} does already exist. Please remove it before indexing again.`")
+            import warnings
+            print(f'\n +---------------------------------------------------------------------------------+ \
+                    \n |                                   🤖🤖🤖                                        | \
+                    \n | The directory workspace already exists. Please remove it before indexing again. | \
```
JoanFM, line 87: I like it, @cristianmtr once you do your example, we could add a reference to the new example in this warning

theUnkownName: Cool, Ill do it for the other examples too then

theUnkownName:
I also refactored all the 'workdir' to have a consistent name across all examples instead of some being name 'tmp_workspace', 
'jina_workspace' and so on.

maximilianwerk:
Sorry to comment that late here, but that refactoring should never have been merged. The `WORKDIR` changes did not happen in all the `.yaml` files and thus this broke several examples. Furthermore, we decided to always have environment variables prefixed with `JINA`.
@nan-wang @JoanFM @theUnkownName 
I'll fix this immediately now.

PR #281, opened 2020-11-12
test: add incremental indexing example

cristianmtr:
Add example of incremental indexing based on South Park

Adapt README

Mention this example as a possible continuation from original South Park.

add integration test

### DIFF
--- a/southpark-incremental-indexing/README.md
+++ b/southpark-incremental-indexing/README.md
@@ -1,0 +1,163 @@
+# Build Bert-based NLP Semantic Search System (with incremental indexing)
+
+<p align="center">
+
+[![Jina](https://github.com/jina-ai/jina/blob/master/.github/badges/jina-badge.svg "We fully commit to open-source")](https://jina.ai)
+[![Jina](https://github.com/jina-ai/jina/blob/master/.github/badges/jina-hello-world-badge.svg "Run Jina 'Hello, World!' without installing anything")](https://github.com/jina-ai/jina#jina-hello-world-)
+[![Jina](https://github.com/jina-ai/jina/blob/master/.github/badges/license-badge.svg "Jina is licensed under Apache-2.0")](#license)
+[![Jina Docs](https://github.com/jina-ai/jina/blob/master/.github/badges/docs-badge.svg "Checkout our docs and learn Jina")](https://docs.jina.ai)
+[![We are hiring](https://github.com/jina-ai/jina/blob/master/.github/badges/jina-corp-badge-hiring.svg "We are hiring full-time position at Jina")](https://jobs.jina.ai)
+<a href="https://twitter.com/intent/tweet?text=%F0%9F%91%8DCheck+out+Jina%3A+the+New+Open-Source+Solution+for+Neural+Information+Retrieval+%F0%9F%94%8D%40JinaAI_&url=https%3A%2F%2Fgithub.com%2Fjina-ai%2Fjina&hashtags=JinaSearch&original_referer=http%3A%2F%2Fgithub.com%2F&tw_p=tweetbutton" target="_blank">
+  <img src="https://github.com/jina-ai/jina/blob/master/.github/badges/twitter-badge.svg"
+       alt="tweet button" title="👍Share Jina with your friends on Twitter"></img>
+</a>
+[![Python 3.7 3.8](https://github.com/jina-ai/jina/blob/master/.github/badges/python-badge.svg "Jina supports Python 3.7 and above")](#)
+[![Docker](https://github.com/jina-ai/jina/blob/master/.github/badges/docker-badge.svg "Jina is multi-arch ready, can run on differnt architectures")](https://hub.docker.com/r/jinaai/jina/tags)
+
+</p>
+
+In this demo, we use Jina to build a semantic search system, with the additional showcase of **incremental indexing**. 
+
+**NOTE**: This example builds on the basic example presented in [here](../southpark-search/README.md). Please study and learn the concepts presented there and then come back to this one.
+
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**
+
+- [Running the example](#running-the-example)
+- [Overview](#overview)
+  - [What is incremental indexing?](#what-is-incremental-indexing)
+  - [Configuration changes](#configuration-changes)
+- [Documentation](#documentation)
+- [Stay tuned](#stay-tuned)
+- [License](#license)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Running the example
+
+**NOTE** Jina only works on Python 3.7 and 3.8
+
+Install requirements.
+
+`pip install -r requirements.txt`
+
+Prepare data
+
+`bash get_data.sh data`
+
+This creates all the data in the `data` dir. Of main import will be `character-lines-1.csv` and `character-lines-2.csv`.
+
+Run the indexing step
+
+`python app.py -t index`
+
+This will perform the indexing on the first file, stop the flow, re-start it, and then index the second. This showcases **incremental indexing**.
+
+Query the data
+
+`python app.py -t query`
+
+
+## Overview
+
+For the discussion on the basic topics, read through the documentation in the [South Park example](../southpark-search/README.md). This README will delve deeper into how we use **incremental indexing**.
+
+### What is incremental indexing?
+
+Briefly, this allows for the user to re-use an index to add new data. It also automatically handles adding documents with duplicate IDs. For a more in-depth explanation on how Jina handles this, refer to [this chapter](https://docs.jina.ai/chapters/prevent_duplicate_indexing/index.html) in our docs.
+
+### Configuration changes
+
+In order to adapt the South Park scripts example to support incremental indexing, we need to to the following:
+
+1. Change the `indexer` in `flow-index` and `flow-query` to use `DocIDCache` as a filter (in the `uses_before` field). This ensures that we prevent duplicates.
+
+    In this example the `DocIDCache` is separated into its own `.yml` file, in `index_cache`:
+    
+    ```yaml
+    !DocIDCache
+    with:
+      index_path: cache.tmp
+    metas:
+      name: cache
+      workspace: $JINA_WORKSPACE
+    requests:
+      on:
+        [SearchRequest, TrainRequest, IndexRequest, ControlRequest]:
+          - !RouteDriver {}
+        IndexRequest:
+          - !TaggingCacheDriver
+            with:
+              tags:
+                is_indexed: true
+          - !FilterQL
+            with:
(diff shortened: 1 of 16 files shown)


southpark-incremental-indexing/README.md
```diff
@@ -0,0 +1,163 @@
...
+
+    # we then re-use the same index to append new data
+    with f:
+        print(f'Indexing file {os.environ["JINA_DATA_FILE_2"]}')
+        f.index_lines(
+           ...
+        )    
+    ```
+    
+    This indexes the first file, closes the flow, and then indexes a second file.
```
JoanFM, line 129: Is it possible to show an example of query showing that it returns documents from both parts?

southpark-incremental-indexing/README.restful.md
```diff
@@ -0,0 +1,259 @@
...
+  - [Run the Docker](#run-the-docker)
+  - [Index and Search By Yourself](#index-and-search-by-yourself)
+  - [Next Steps](#next-steps)
+  - [Documentation](#documentation)
+  - [Stay tuned](#stay-tuned)
+  - [License](#license)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Jina ❤️ Docker and RESTful APIs
```
maximilianwerk, line 15: Do we really need `README.restful.md`? I feel like that is not the point of this demo.

maximilianwerk (commented):
I am twisted between liking to have the whole southpark C&P here and just add another `app-incremental.py` to the original southpark example. In case you have not discussed this beforehand with @alexcg1, could we get his view on this? Otherwise, I like the simplicity of just using the index flow twice. That's how it should work.

southpark-incremental-indexing/README.restful.md
cristianmtr: I agree. Removed

alexcg1: I think we should NOT c&p the whole South Park README. We should:

1. Say "do the South Park example first"
2. Explain "now you have the South Park code, here's how you modify it for incremental indexing"

This 
- makes maintenance easier
- reduces workload
- in case someone has done south park already and wants to add incremental indexing, they won't need to hunt around WHERE that section is in this new README

alexcg1: As we move forwards into restructuring the tutorials, we'll break them down into:

- fundamental unit (e.g. south park search) (i.e. the most basic Jina thing you can program that does sth useful)
- fundamental unit + 1 thing - e.g. south park + incremental indexing
- fundamental unit + 1 different thing - e.g. south park + QueryLang

Since each of these would have different pods and flows, I'd suggest putting them in their own folders (for now), prefixed with `southpark_` (or whatever)

cristianmtr: Yes, that's how I've done the main README.md

Max's comment was about the .restful

southpark-incremental-indexing/README.md
cristianmtr: It's in the integration test now

southpark-incremental-indexing/README.restful.md
alexcg1: D'oh. Hang on, I'll check that too

alexcg1: Yeah, I vote for removing the restful README

JoanFM:
The point is if besides the `README.md` we can also copy & paste code or we only want to have a document explaining the changes? I prefer to have some code, but we could maybe avoid adding the `get_data`, `prepare_data`... and so on?

cristianmtr:
> The point is if besides the `README.md` we can also copy & paste code or we only want to have a document explaining the changes? I prefer to have some code, but we could maybe avoid adding the `get_data`, `prepare_data`... and so on?

I guess we could add a script that copies the required files, perhaps? (`get_data` etc.)

JoanFM:
can we check if it works when running the `get_data ...` from `southpark`?

cristianmtr:
> can we check if it works when running the `get_data ...` from `southpark`?

It works but we still need to adapt the `prep_data` file, since in the incremental indexing example we want to split it into two `.csv` files. So would need to write that one

alexcg1:
Agreed with @JoanFM on `get_data` etc

Looking at the README, the code changes seem very simple:

- For a tutorial (something you learn from), I would say just keep the README and throw everything else away since it's not needed if you want to implement incremental indexing on your own search
- However, for a self-contained example (i.e. something you can run) we would need to keep everything needed to run `app.py` successfully

This is one of the bigger issues we have with our docs - and something we're trying to untangle.

It's not an easy call for me (since I love tutorials a lot more) but I'm gonna go with keep anything needed to run `app.py` successfully. That's the example itself. The readme is (more or less) the tutorial part

JoanFM (approved):

JoanFM (changes requested):
Remove the inner `.github` subfolder inside this example?

southpark-incremental-indexing/README.md
```diff
@@ -0,0 +1,163 @@
...
+          - !FilterQL
+            with:
+              lookups: {tags__is_indexed__neq: true}
+    ```
+    
+    This might look complicated, but what it basically does is to first check the cache for any matching doc IDs, before indexing and querying.
+
+2. Adapt dataset to highlight incremental indexing.
+
+    We split the dataset into two files, `character-lines-1.csv` and `characters-lines-2.csv`. This way we can show how we can index one, close the `Flow` object, and then index the other.
```
JoanFM, line 104: Is there an example where we can show that the results come from the 2 different files? or that the size of the `index` is the expected at every time?

I would even add the showcase that if you want to `index` "character-lines-1.csv" a second time, nothing happens

southpark-incremental-indexing/tests/test_southparksearch.py
```diff
@@ -0,0 +1,79 @@
...
+def get_flow():
+    f = Flow().load_config(QUERY_FLOW_FILE_PATH)
+    f.use_rest_gateway()
+    return f
+
+
+@pytest.fixture
+def queries():
+    return [("Don't say anything\n",
+             ["Don't say anything\n", "Check that: I'll watch that game.\n", "I'm trying to talk\n"]),
```
JoanFM, line 61: Do the results come from different sources?

cristianmtr:
> Remove the inner `.github` subfolder inside this example?

Done

southpark-incremental-indexing/README.md
cristianmtr: Good idea. It's been added in the integration test.

southpark-incremental-indexing/tests/test_southparksearch.py
cristianmtr: Yep

southpark-incremental-indexing/tests/distributed/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN git clone https://github.com/jina-ai/jinad.git && \
 
 RUN python -c "from transformers import DistilBertModel, DistilBertTokenizer; x='distilbert-base-cased'; DistilBertModel.from_pretrained(x); DistilBertTokenizer.from_pretrained(x)"
 
-COPY . /
```
JoanFM, line 19: is this being moved? I do not like this so strange workdir reference. Where is this file?

southpark-incremental-indexing/tests/distributed/docker-compose.yml
```diff
@@ -3,7 +3,7 @@ services:
   flow:
     image: south
     build:
-      context: .
+      context: ../../../southpark-incremental-indexing/tests/distributed
```
JoanFM, line 6: is it an intention to adapt the old test with the `incremental indexing` one? I would keep it as it was? I'd try to avoid having this dependency

JoanFM (commented):
I am a little bit confused by the workdir of the docker tests

southpark-incremental-indexing/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN python -c "from transformers import DistilBertModel, DistilBertTokenizer; x='distilbert-base-cased'; DistilBertModel.from_pretrained(x); DistilBertTokenizer.from_pretrained(x)"
+
+COPY . /
+
+RUN bash get_data.sh ./data && \
+    python app.py -t index && \
+    rm -rf data
+
+ENTRYPOINT ["python", "app.py", "-t", "query_restful"]
```
maximilianwerk, line 21: We don't need any Dockerfile for this example. I'd remove it.

maximilianwerk (changes requested):
Why are the distributed files copied here? Shouldn't they stay in the other southpark?

Furthermore, I'd leave docker out of the game for now. Or do we need this for some integration test running?

cristianmtr:
> Why are the distributed files copied here? Shouldn't they stay in the other southpark?
> 
> Furthermore, I'd leave docker out of the game for now. Or do we need this for some integration test running?

My understanding was we were planning on adding this as an integration test in CI, right?

southpark-incremental-indexing/tests/distributed/docker-compose.yml
cristianmtr: Yeah, that's what I thought initially. But if we don't want that, I can remove it (or maybe we could add it later as a separate PR to keep this one to the point)

JoanFM: let's remove it for now and maybe add in a separate PR if needed

southpark-incremental-indexing/app.py
```diff
@@ -0,0 +1,112 @@
...
+    with f:
+        f.dry_run()
+
+
+@click.command()
+@click.option(
+    "--task",
+    "-t",
+    type=click.Choice(
+        ["index", "query", "query_restful", "dryrun"], case_sensitive=False
```
JoanFM, line 89: I think `dryrun` is deprecated now, but no worries

southpark-incremental-indexing/app.py
```diff
@@ -0,0 +1,112 @@
...
+
+import click
+from jina.flow import Flow
+
+
+def config():
+    os.environ["JINA_DATA_FILE_1"] = os.environ.get(
+        "JINA_DATA_FILE_1", "data/character-lines-1.csv"
+    )
+    os.environ["JINA_DATA_FILE_2"] = os.environ.get(
```
JoanFM, line 14: would be nice to migrate this and southpark to single quote

cristianmtr:
Removed Dockerfile

JoanFM (approved):
Some minor things but LGTM!

southpark-incremental-indexing/app.py
cristianmtr: removed

southpark-incremental-indexing/app.py
cristianmtr: good point. done (for this example). cleaning up double quotes throughout the others should be a separate PR, I think

JoanFM: Sure

maximilianwerk (approved):

PR #277, opened 2020-11-10
feat: add terraform script for southpark

rutujasurve94:
- Adds `terraform` resources for creation of `aws` service, tasks, cluster, ecs repository
- Adds instances for `index`, `encode` and `flow` phases.

PR #284, opened 2020-11-13
feat: terraform case study - ecs for flow, encoder, indexer

theUnkownName:


rutujasurve94 (changes requested):
Let's update this PR wrt #277's approach:
Previously we tried to spin up 3 EC2 instances instead of ECS (Flow, encoder, indexer) without Fargate. However, we decided to go back to Fargate because we were facing many problems trying to manage all the resources without Fargate as it was already working for NON distributed Southpark example last week. So right now the terraform file we have can do the following:

Set one VPC
As we are using Fargate, we have ECS instead of EC2, an ECS needs a Task, a Service and a Cluster, all of them are also all spun with the current .tf file;
For this Task we made 3 containers, a Flow, an Encoder and an Indexer. And they have the docker image attached to it.
The Service will run this task(s) and this is the one that has the Load Balancer. The Load Balancer is listening to the port 45678
The Cluster will group all of those things together
Now, being that said, even tho the services are running in AWS, when I try to curl (curl --request POST -d '{"top_k": 10, "mode": "search", "data": ["text:hey, dude"]}' -H 'Content-Type: application/json' 'http://southpark-lb-tf-1350216710.us-east-2.elb.amazonaws.com' ) it's not working and I'm unsure on what could be wrong at this point and it also seems we've been using a lot of time with this.
Here are the slides with the structure we had in mind for the 2 approches (Fargate no Fargate) https://docs.google.com/presentation/d/1LakYEv7LMBBV4g9Cop8-cekawsaZhNN4kEjFjeSrASQ/edit#slide=id.gac19b5f751_0_34


Let's update this PR with the above approach for our case study

nan-wang:
What's the difference between `tests/distributed/` and `tests/distributed/terraform_simple`?

theUnkownName:
Should we keep this @maximilianwerk ? this was the version without Fargate but since we decided to go with it (Fargate)  I don't know if it makes sense to keep this one too

maximilianwerk:
> Should we keep this @maximilianwerk ? this was the version without Fargate but since we decided to go with it (Fargate) I don't know if it makes sense to keep this one too

I think, that depends on the answer to the following question: Is there sufficient knowledge inside this PR, that we definitely will need to leverage in the future and you need it as a look-up draft? If so, let's move the code somewhere, were we can read it up later (perhaps cloud-ops repo. not sure). If you can get this knowledge also easily from other information sources that you read, let's close this PR and throw it away.

theUnkownName:
> > Should we keep this @maximilianwerk ? this was the version without Fargate but since we decided to go with it (Fargate) I don't know if it makes sense to keep this one too
> 
> I think, that depends on the answer to the following question: Is there sufficient knowledge inside this PR, that we definitely will need to leverage in the future and you need it as a look-up draft? If so, let's move the code somewhere, were we can read it up later (perhaps cloud-ops repo. not sure). If you can get this knowledge also easily from other information sources that you read, let's close this PR and throw it away.

It has some research into it, but I could do it again if needed and also, it would probably look cleaner next time because it wouldn't be as messy. And plus that, I have my local copy, so I would suggest to close it.

PR #287, opened 2020-11-14
test: adapt proper testing

bwanglzu:
1. Adapt image encoder and multimodal encoder with fashion200k dataset.
2. Add tests and expected npy.

Need to update `app.py` with default parameters.


multimodal-search-tirg/TirgImageEncoder/__init__.py
```diff
@@ -37,7 +37,7 @@ def post_init(self):
                 texts = pickle.load(fp)
             self.model = TIRG(texts, 512)
             model_sd = torch.load(self.model_path, map_location=torch.device('cpu'))
-            self.model.load_state_dict(model_sd['model_state_dict'])
+            self.model.load_state_dict(model_sd['model_state_dict'], strict=False)
```
JoanFM, line 40: what does it do? just curious?

multimodal-search-tirg/TirgMultimodalEncoder/Dockerfile
```diff
@@ -7,13 +7,10 @@ WORKDIR /workspace
...
     && apt-get -y install wget \
-    && wget https://storage.googleapis.com/image_retrieval_css/pretrained_models/checkpoint_fashion200k.pth \
-    && wget https://github.com/bhavsarpratik/tirg/raw/master/texts.pkl
+    && wget https://storage.googleapis.com/image_retrieval_css/pretrained_models/checkpoint_css3d.pth \
+    && wget https://github.com/bwanglzu/tirg-css/raw/main/texts.pkl
 
 # install the third-party requirements
 RUN pip install -r requirements.txt
 
-# for testing the image
```
JoanFM, line 16: remember to add back the tests

multimodal-search-tirg/TirgImageEncoder/__init__.py
bwanglzu: @JoanFM I uncounted the same error as described [here](https://discuss.pytorch.org/t/missing-keys-unexpected-keys-in-state-dict-when-loading-self-trained-model/22379) while initializing pretrained tirg model.

bhavsarpratik: We were not facing this problem in the fashion model. Not sure why is it coming in CSS model. What happens when it tried to load weight from the state_dict but doesn't find the key. Does it initialise randomly? The documentation says 
"Whether you are loading from a partial state_dict, which is missing some keys, or loading a state_dict with more keys than the model that you are loading into, you can set the strict argument to False in the load_state_dict() function to ignore non-matching keys."

multimodal-search-tirg/TirgMultimodalEncoder/__init__.py
```diff
@@ -56,8 +56,7 @@ def _get_features(self, data):
             visual_data = np.moveaxis(visual_data, self.channel_axis, self._default_channel_axis)
         textual_data = data[(self.positional_modality.index('text'))]
 
-        visual_data = torch.from_numpy(np.stack(visual_data)).float()
-        textual_data = np.stack(textual_data).tolist()
+        visual_data = torch.stack(visual_data).float()
```
bhavsarpratik, line 59: In one of my earlier push I had to change below make flow work along with *data in argument and `data` in encode instead of `*data`
From
```
visual_data = torch.stack(visual_data).float()
```

to
```
visual_data = torch.from_numpy(np.stack(visual_data)).float()
textual_data = np.stack(textual_data).tolist()
```

multimodal-search-tirg/TirgImageEncoder/Dockerfile
```diff
@@ -7,8 +7,8 @@ WORKDIR /workspace
 # Get model and texts file
 RUN apt-get -y update \
     && apt-get -y install wget \
-    && wget https://storage.googleapis.com/image_retrieval_css/pretrained_models/checkpoint_fashion200k.pth \
-    && wget https://github.com/bhavsarpratik/tirg/raw/master/texts.pkl
+    && wget https://storage.googleapis.com/image_retrieval_css/pretrained_models/checkpoint_css3d.pth \
+    && wget https://github.com/bwanglzu/tirg-css/raw/main/texts.pkl
```
bhavsarpratik, line 11: Lets rename the model to `checkpoint.pth` so we don't have to change its name everywhere in code

bhavsarpratik (changes requested):

JoanFM:
@bhavsarpratik I think is good for merging, then you can have any tweak u want

PR #289, opened 2020-11-17
feat: add evaluate and train example

bhavsarpratik:
DRAFT: This is combined PR of fashion-mnist eval and hypertune/train.


JoanFM:
Just as a comment, I would separate the evaluation example and the hypetrain right? I do not think it will become an example the `hypertrain` now

fashion-hyperparameter/optimize.py
```diff
@@ -0,0 +1,210 @@
...
+                'JINA_INDEX_DATA_FILE': 'tests/hyperparameter/index_data.csv',
+                'JINA_EVALUATION_DATA_FILE': 'tests/hyperparameter/query_data.csv',
+                'JINA_WORKSPACE': 'workspace_eval',
+                'JINA_MATRIX_SIZE': size,
+            }
+            for size in range(10, 100)
+        ]
+
+    def parameters_iterator(self):
+        yield {}, self.save_results
```
JoanFM, line 24: it should not get the next `parameter`? how does this work?

bhavsarpratik:
Yes we just discussed it. We will go with both for now and later separate out things or even move some to jina core and refactor.

fashion-hyperparameter/optimize.py
maximilianwerk: Please ignore this code for now. We are still implementing and haven't deleted the parts, which were just sketches. This draft is for us to see better what is already added.

fashion-hyperparameter/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+COPY tests/hyperparameter/requirements.txt /tests/hyperparameter/requirements.txt
+
+RUN python -m pip install --no-cache-dir --upgrade pip && \
+    pip install --no-cache-dir -r tests/hyperparameter/requirements.txt
+
+RUN git clone https://github.com/jina-ai/jinad.git && \
+    pip install --no-cache-dir -r jinad/requirements.txt
+
+RUN python -c "from transformers import DistilBertModel, DistilBertTokenizer; x='distilbert-base-cased'; DistilBertModel.from_pretrained(x); DistilBertTokenizer.from_pretrained(x)"
```
nan-wang, line 17:
```suggestion

```

bhavsarpratik:
`pip install optuna`
Evaluate 
`python evaluate.py`
Hypertune 
`python optimize.py`

fashion-hyperparameter/optimize_class.py
```diff
@@ -0,0 +1,106 @@
...
+        self.query_batch_size = query_batch_size
+        self.trial_parameter_sampler = trial_parameter_sampler
+        self.n_trials = n_trials
+        self.direction = direction
+        self.seed = seed
+        self.config_dir = config_dir
+        self.best_config_filename = best_config_filename
+
+
+    def run_indexing(self):
```
JoanFM, line 33: Just a note for future design, evaluation is designed to be `flow` agnostic (in theory one can evaluate any flow (even an Index one)). So the final version should not assume that we are running `indexing` and later `querying`. I think this `indexing` part should be like a `pre_hook` callback or something.

fashion-hyperparameter/optimize_class.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @staticmethod
+    def parameters_to_env(parameters):
+        for environment_variable, value in parameters.items():
+            os.environ[environment_variable] = str(value)
+
+    def run_evaluation(self, parameters, evaluation_callback):
+        Optimizer.parameters_to_env(parameters)
+        self.run_indexing()
+        self.run_querying(evaluation_callback)
```
JoanFM, line 63: so for instance, you could do `self.run(pre_hook=self.index(), post_hook=evaluation_callback())` or something in these lines

fashion-hyperparameter/optimize_class.py
bhavsarpratik: This idea might be too advanced for our users. Not sure about it. We can rename the `run_querying` to `run_evaluation`

fashion-hyperparameter/optimize_class.py
bhavsarpratik: Does there exist any situation where pre_hook is different than index?

bhavsarpratik: Does there exist any situation where post_hook is different than eval?

JoanFM: It could be empty. 

I am not meaning to put it in the example. But when designing the module classes to have it in mind. Another thing is how u expose these features to the user.

JoanFM: for post_hook different than eval it may make sense to run evaluation by default yes. 

I am just raising awareness for the definitive design to keep that in mind. We should be coherent with the idea that `evaluation` is agnostic to the `flow` we are using

maximilianwerk: Perhaps we should just give in a list of flows with there respective evaluation callbacks. Then we guarantee full flexibility.

JoanFM: Yes, but would be important to have different 'pre-processing' (here run an index flow) per each evaluation? Or this preprocessing could be shared by all the Flows in the Pool?

bhavsarpratik:
@maximilianwerk The logger is still printing everything. Need to probe more on this

fashion-hyperparameter/optimize_class.py
```diff
@@ -0,0 +1,204 @@
...
+        self.query_document_generator, query_document_generator = tee(self.query_document_generator)
+
+        with Flow().load_config(query_yaml) as f:
+            f.search(
+                query_document_generator,
+                batch_size=self.query_batch_size,
+                output_fn=callback
+            )
+
+    def run_evaluation(self, index_yaml, query_yaml, workspace, evaluation_callback):
```
JoanFM, line 78: What if I want to evaluate an indexing flow? Remember that evaluation is agnostic to the flow u run. One can evaluate anything

PR #291, opened 2020-11-17
feat: add shards and parallel

bhavsarpratik:


multimodal-search-tirg/flow-index.yml
```diff
@@ -5,9 +5,11 @@ pods:
   read:
     uses: pods/image-read.yml
     read_only: true
+    parallel: $PARALLEL
```
JoanFM, line 8: I would avoid the parallel here

multimodal-search-tirg/flow-index.yml
```diff
@@ -5,9 +5,11 @@ pods:
...
   normalize:
     uses: pods/normalize.yml
     read_only: true
+    parallel: $PARALLEL
```
JoanFM, line 12: and also here

multimodal-search-tirg/flow-index.yml
```diff
@@ -5,9 +5,11 @@ pods:
...
   encode:
     uses: jinaai/hub.executors.encoders.image.tirg
```
JoanFM, line 14: there is no parallelization here, why is it failing here?

bhavsarpratik:
@JoanFM Please check and merge for your test

PR #295, opened 2020-11-19
fix: remove laser encoder module

bhavsarpratik:
Removing as now we have native support for LaserEncoder


JoanFM:
I think this should be reverted, because multilingual does not work with the jina version for which LaserEncoder is added, or did u test it with the version in `requirements.txt`?

PR #296, opened 2020-11-19
chore: added multimodal to main readme

bhavsarpratik:


README.md
```diff
@@ -209,7 +209,26 @@ Use prefetching and sharding to improve the performance of your index and query
...
+
+<table width=100% style='width: 100%'>
+<tr>
+<th width="90%" style="width: 90%">Tutorials</th>
+<th width="10%" style="width: 10%">Level</th>
+</tr>
+
+<tr>
+<td>
+<h4><a href="https://github.com/jina-ai/examples/tree/master/multimodal-search-tirg">Search images with 2 modalities in the query</a></h4>
```
JoanFM, line 227: Should we say search images with transformations given by short descriptions?

alexcg1: Let's keep it as simple as we can without making it any simpler. Since this readme is the first thing people see, and aims to be:
- a jumping off point for tutorials
- a way to show our capabilities (which may not be obvious if we use jargon when simple words would do)

PR #301, opened 2020-11-25
fix: added uri conversion for jinabox

bhavsarpratik:
For multimodal in jinabox

### DIFF
--- a/multimodal-search-tirg/pods/image-read.yml
+++ b/multimodal-search-tirg/pods/image-read.yml
@@ -10,7 +10,9 @@ requests:
         with:
           traversal_paths: ['c']
           lookups: {'modality': 'image'}
-      # - !URI2Buffer {}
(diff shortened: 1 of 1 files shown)


JoanFM (commented):
Isn't there a `Buffer2Uri` before the `KVIndexer`?

bhavsarpratik:
In object-search there is one `Buffer2Uri` in doc.yml's index side which we also have. Although we need to worry only searchrequest side

JoanFM:
> In object-search there is one `Buffer2Uri` in doc.yml's index side which we also have. Although we need to worry only searchrequest side

So right now we are using some uri parsing at the client side to show the results?

bhavsarpratik:
Yes we have conversion in our `app.py`

JoanFM:
> Yes we have conversion in our `app.py`

Ok now I understand!

PR #302, opened 2020-11-25
Feat add cloudwatch

theUnkownName:


rutujasurve94 (approved):

southpark-search/tests/distributed/terraform.tf
```diff
@@ -107,6 +91,16 @@ data "aws_iam_policy_document" "assume_role_policy" {
   }
 }
 
+resource "aws_cloudwatch_log_group" "southpark_loggroup" {
+  name              = "southpark-log"
+  retention_in_days = 1
+}
+
+resource "aws_cloudwatch_log_stream" "southpark-distributed-logs" {
+  name           = "southpark-distributed-logs"
```
rutujasurve94, line 100: Do we plan on using this log stream of distributed logs?

rutujasurve94:
LGTM, just a comment on use of distributed log stream.
We could even get a sample log link after terraform apply and later link it to the `README`

southpark-search/tests/distributed/terraform.tf
theUnkownName: Yeah I think it would make sense or?

PR #304, opened 2020-11-30
refactor: refactor old array2pb into new data primitive type

theUnkownName:


fashion-example-query/fashion/app.py
```diff
@@ -149,13 +150,15 @@ def query(num_doc, target: dict):
...
     parallel = 2 if task == 'index' else 1
+    shards = 1
     os.environ['RESOURCE_DIR'] = resource_filename('jina', 'resources')
-    os.environ['SHARDS'] = str(1)
+    os.environ['SHARDS'] = str(shards)
     os.environ['PARALLEL'] = str(parallel)
-    os.environ['WORKDIR'] = './workspace'
-    os.makedirs(os.environ['WORKDIR'], exist_ok=True)
+    os.environ['HW_WORKDIR'] = './workspace'
+    os.makedirs(os.environ['HW_WORKDIR'], exist_ok=True)
```
maximilianwerk, line 158: Since we anyhow touch the environment variables, can we please prefix them with `JINA_`? Thanks a lot.

theUnkownName: Sure, makes sense

theUnkownName: Done

maximilianwerk (approved):
LGTM

fashion-example-query/fashion/app.py
```diff
@@ -2,21 +2,21 @@
 __license__ = "Apache-2.0"
 
 import os
-import click
+import click
```
cristianmtr, line 5: I'm guessing these linting issues will be taken care of by jina bot?

fashion-example-query/fashion/app.py
```diff
@@ -114,23 +113,25 @@ def download_data(target: dict, download_proxy=None):
                 v['data'] = load_mnist(v['filename'])
 
 
-def index_generator(num_doc, target: dict):
+def index_generator(num_doc: int, target: dict):
     for j in range(num_doc):
+        d = Document(content=target['index']['data'][j])
```
cristianmtr, line 118: I think the pattern is to do

```
with Document(...) as d:
   d.content = 
```

and then the `update_id` is done by the context managing,no?

fashion-example-query/fashion/app.py
```diff
@@ -204,4 +204,4 @@ def main(task, num_docs_query, num_docs_index):
 
 
 if __name__ == '__main__':
-    main()
+    main()
```
cristianmtr, line 207: Same here for linting. EOF should have `\n`

cristianmtr (changes requested):
LGTM but some questions

fashion-example-query/fashion/app.py
theUnkownName: Since we were talking about maybe not having this context manager anymore, I thought to keep it like this.

PR #305, opened 2020-11-30
refactor(southparksearch): updating version numbers

florian-hoenicke:
setting version numbers to have deterministic builds


JoanFM:
Not valid anymore as it is already done

PR #313, opened 2020-12-06
older version of jina is not compatible with the newer version of doc…

prabhupad26:
jina v0.4.1 is not compatible with the docker 4.4.0 , downgrading the docker sdk to 4.3.0 seems to be working fine.


JoanFM:
recheckcla

PR #314, opened 2020-12-07
fix(flower-search): updated url for 17flowers data set

serge-m:
updated url for 17flowers data set. Otherwise get_data fails.

UPD:
The old URL works, but it downloads a redirect page, then tar command fails.

```
$ curl http://www.robots.ox.ac.uk/~vgg/data/flowers/17/17flowers.tgz
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="https://www.robots.ox.ac.uk/~vgg/data/flowers/17/17flowers.tgz">here</a>.</p>
</body></html>
```

### DIFF
--- a/flower-search/get_data.sh
+++ b/flower-search/get_data.sh
@@ -2,5 +2,5 @@
 TEST_WORKDIR=/tmp/jina/flower/
 mkdir -p ${TEST_WORKDIR}
 cd ${TEST_WORKDIR}
-curl http://www.robots.ox.ac.uk/~vgg/data/flowers/17/17flowers.tgz --output 17flowers.tgz
+curl https://www.robots.ox.ac.uk/~vgg/data/flowers/17/17flowers.tgz --output 17flowers.tgz
 tar zxf 17flowers.tgz


serge-m:
test failure seems to be related to https://github.com/jina-ai/examples/issues/276

cristianmtr:
> test failure seems to be related to #276

Yes, I have re-run the checks now. It shouldn't fail

flower-search/get_data.sh
```diff
@@ -2,5 +2,5 @@
 TEST_WORKDIR=/tmp/jina/flower/
 mkdir -p ${TEST_WORKDIR}
 cd ${TEST_WORKDIR}
-curl http://www.robots.ox.ac.uk/~vgg/data/flowers/17/17flowers.tgz --output 17flowers.tgz
+curl https://www.robots.ox.ac.uk/~vgg/data/flowers/17/17flowers.tgz --output 17flowers.tgz
```
cristianmtr, line 5: For me the old one still works. But I agree `https` is preferable when available

cristianmtr (approved):

flower-search/get_data.sh
serge-m: For me the old one works, but it downloads a redirect page, then `tar` command fails. Does it really download the correct archive for you?

Alternatively one could add -L flag to curl

cristianmtr: Oh I see. I only tested it in the browser.

alexcg1:
Can someone merge this please? It's failing a test for multires-lyrics search, yet it doesn't touch that code at all. From my testing over past week so many of our examples have issues that tests will fail regardless of whether the PR code is good or not

PR #318, opened 2020-12-11
fix: adapt faiss search

JoanFM:
Trial to update faiss search to new version.

Problems: Waiting for https://github.com/jina-ai/jina/issues/1449 to be solved

Extra problems: It is very confusing to map the proper volumes so that they can be used in the workspace of the Faiss Container

Extra problems: Cannot use ref_indexer solution as described in https://github.com/jina-ai/jina/issues/1438


ivana-bt:
Hi @JoanFM ,

thanks for the refactoring! I tried the new faiss search logic, but it seems that the `id` from match is now a hash instead of an id whereas the groundtruth data returns an actual id :thinking: how is it possible to get a "real" id (I'm using a recent `master` version of jina).

Many thanks!

JoanFM:
> Hi @JoanFM ,
> 
> thanks for the refactoring! I tried the new faiss search logic, but it seems that the `id` from match is now a hash instead of an id whereas the groundtruth data returns an actual id  how is it possible to get a "real" id (I'm using a recent `master` version of jina).
> 
> Many thanks!

Hey @ivana-bt ,

Would you open an issue describing the problem in more detail?

JoanFM:
> Hi @JoanFM ,
> 
> thanks for the refactoring! I tried the new faiss search logic, but it seems that the `id` from match is now a hash instead of an id whereas the groundtruth data returns an actual id  how is it possible to get a "real" id (I'm using a recent `master` version of jina).
> 
> Many thanks!

Hey @ivana-bt ,

I recently had time to work on some fixes for this example, and I added some changes to it

advanced-vector-search/flow-index.yml
```diff
@@ -6,9 +6,8 @@ pods:
     uses: _forward
```
theUnkownName, line 6: We still need this forward? I remember we had it to make it consistent with the other examples we had but is still that the case?

JoanFM: it is not needed, we can remove

advanced-vector-search/README.md
```diff
@@ -46,19 +56,20 @@ Before moving forward, we highly suggest completing/reviewing our lovely [Jina 1
...
 
-pip install -r requirements.txt
+```bash
+./get_siftsmall.sh
 ```
 
-Additionally, since FAISS introduces many dependencies, the [Jina hub image](https://github.com/jina-ai/jina-hub/tree/master/indexers/vector/FaissIndexer) is used for simplicity.
-
-## Prepare the data
+Moreover,
```
theUnkownName, line 72: It has a weird format, maybe a new line was added?

advanced-vector-search/README.md
```diff
@@ -170,79 +174,48 @@ python app.py -t index -n $batch_size
...
+Recall@100 => 77.69999986886978%
+```
 
-Using more complex inverted indices and encoders (i.e., different `index_key`) should lead to better results. Using `index_key: 'Flat'` gives a recall equal to 1 because it is the exhaustive search mode for FAISS. To learn more about the range of supported keys and options, be sure to visit [FAISS indexes](https://github.com/facebookresearch/faiss/wiki/Faiss-indexes).
+It can be good to look for different parameters to guarantee the best results
 
 ## Wrap up
 
-In this example we have seen how to use `FaissIndexer` to use FAISS as a vector database. We also have seen how to use a pod inside a docker container inside our index and query flows.
+In this example we have seen how to ude different indexers as vector databases and how to use a `ref_indexer` as a base indexer.
```
theUnkownName, line 210: Little typo "use"

advanced-vector-search/README.md
JoanFM: well there is a line between Moreover and Faiss, maybe we can refactor later

advanced-vector-search/README.md
JoanFM: can u add suggestion so that I can commit here?

deepankarm (approved):

advanced-vector-search/README.md
```diff
@@ -120,7 +122,8 @@ pods:
 
 ### Query <!-- omit in toc -->
 
-Just as we need to index, we also need a Flow to process the request message during querying. The only difference with its index counterpart is that `doc_indexer` is piped to the `faiss_indexer` (see below).
+Just as we need to index, we also need a Flow to process the request message during querying. The query flow looks very similar
```
cristianmtr, line 125: Why is there a line break here?

advanced-vector-search/README.md
```diff
@@ -170,79 +172,48 @@ python app.py -t index -n $batch_size
 
 ### Query <!-- omit in toc -->
 
-Query can be run with the following command.
+Query can be run with the following command, where index_type can be `annoy`, `faiss`, or `numpy`.
```
cristianmtr, line 175: I'd say no line break here

PR #355, opened 2021-01-18
refactor: update example

theUnkownName:
- [x] Reduce to 2-3 indexes
- [x] Update to 0.9 
- [x] Switch to Jina types 
- [x] Update readme
- [ ] Update copyrights


fashion-example-query/app.py
```diff
@@ -116,7 +103,6 @@ def download_data(target: dict, download_proxy=None):
 def index_generator(num_doc: int, target: dict):
     for j in range(num_doc):
         d = Document(content=target['index']['data'][j])
```
JoanFM, line 105: try to use `document` as context manager

fashion-example-query/app.py
```diff
@@ -116,7 +103,6 @@ def download_data(target: dict, download_proxy=None):
...
-        d.update_id()
         label_int = target['index-labels']['data'][j][0]
```
JoanFM, line 106: maybe u should avoid indexing images of a label not belonging to the labels?

JoanFM (changes requested):

fashion-example-query/README.md
```diff
@@ -43,31 +43,19 @@ python app.py index
 First up we need to build up an index and then we'll search through this index when we use the query Flow later.
 So what is happening here?
 
-There are 10 categories in the fashion-mnist-data
+There are 10 categories in the fashion-mnist-data, but to simplify this example we will only use 3 categories:
 
     0	        T-shirt/top
     1	        Trouser
     2	        Pullover
```
cristianmtr, line 50: Any reason why we're sticking to `Pullover`, and not `pullover`?

fashion-example-query/README.md
```diff
@@ -84,21 +72,21 @@ python app.py query
...
+So if we run this as it is, we will see our results filtered by Pullovers like this:
 
 ![alt text](.github/images/results.png "Results")
 
 
 ## QueryLanguage
 
 So the real magic here, if you compare it with the original HelloWorld, is that we are using QueryLanguage.
-This is happening in the index and query flow of each category, for example, for the ```indexer-dress.yml``` we have
+This is happening in the index and query flow of each category, for example, for the ```indexer-Pullover.yml``` we have
```
cristianmtr, line 83: It's kinda odd in filenames, no?

fashion-example-query/flows/index.yml
```diff
@@ -0,0 +1,29 @@
...
+    shards: $JINA_SHARDS
+    separated_workspace: true
+    needs: [encode]
+  - name: indexer_trouser
+    uses: pods/indexer-trouser.yml
+    shards: $JINA_SHARDS
+    separated_workspace: true
+    needs: [encode]
+  - name: indexer_pullover
+    uses: pods/indexer-pullover.yml
```
cristianmtr, line 22: Ah, I see it is lowercase after all. So maybe make it consistent in the README

fashion-example-query/requirements.txt
```diff
@@ -1,2 +1,6 @@
-jina[devel,torch,tensorflow,transformers]==0.7.12
 click
```
cristianmtr, line 1: Specific version

cristianmtr (changes requested):
some changes

fashion-example-query/requirements.txt
theUnkownName: I think I already changed this on the latest commit

cristianmtr: I don't see it for `click`.

fashion-example-query/app.py
```diff
@@ -61,7 +49,7 @@ def print_result(resp):
         result_html.append(f'<tr><td><img src="{vi}"/></td><td>')
         for kk in d.matches:
```
Yongxuanzhang, line 50: I think the variables within print_result() is not very readable. e.g. Shall we change d, vi, kk to more readable names?

Yongxuanzhang (commented):
I think the variables within print_result() is not very readable. e.g. Shall we change d, vi, kk to more readable names?

fashion-example-query/app.py
JoanFM: if u have a way to do so, u can also do that in the `hello-world` in the core for sure

Yongxuanzhang: I'm not sure what d, vi, kk, kmi represent. I would prefer some more descriptive ones, but not sure how to rename them.
And I also feel like 'd' is not a good naming convention. For example, if we use pdb to debug here, 'd' won't work. (https://stackoverflow.com/questions/15160946/python-debugger-tells-me-value-of-numpy-array-is-newest-frame).
What do you think?

JoanFM: vi is `uri` and kk is `match`

Yongxuanzhang: Yeah. I mean what are the full names of these abbreviations. 😄  
My thought is 
- d->data or query_data
- vi->data_uri or query_data_uri
- kk->match
- kmi->match_uri

JoanFM: Good lets go

theUnkownName: Sounds good, so we change it in this example and in the hello-workd example in core right?

fashion-example-query/flows/index.yml
```diff
@@ -0,0 +1,29 @@
+!Flow
+version: '1'
+with:
```
JoanFM, line 3: remove logserver option as it has been deprecated

JoanFM (commented):
can u try to update to 0.9.20 (latest release) and removing the separated workspace parameter?

fashion-example-query/flows/query.yml
```diff
@@ -0,0 +1,34 @@
...
+pods:
+  - name: encode
+    uses: pods/encode.yml
+    parallel: $JINA_PARALLEL
+  - name: indexer_tshirt
+    uses: pods/indexer-tshirt.yml
+    shards: $JINA_SHARDS
+    needs: [encode]
+    polling: all
+    uses_after: pods/reduce.yml
```
JoanFM, line 15: make sure `reduce.yml` do proper sorting of results

JoanFM (changes requested):

fashion-example-query/flows/query.yml
theUnkownName: But this is the same as before no? It was/is already giving good results

fashion-example-query/flows/index.yml
```diff
@@ -0,0 +1,25 @@
...
+  compress_hwm: 1024
+pods:
+  - name: encode
+    uses: pods/encode.yml
+    parallel: $JINA_PARALLEL
```
cristianmtr, line 8: make sure you consistently use `shards` (`shards: $JINA_SHARDS_ENCOER` etc.)

cristianmtr (changes requested):

fashion-example-query/flows/index.yml
theUnkownName: By this you mean that in general we will use `parallel` as `JINA_SHARDS_ENCOER`?

fashion-example-query/flows/query.yml
JoanFM: Yes I checked and is good

fashion-example-query/flows/index.yml
cristianmtr: I meant:

1. Let's try to always use `shards` (and `parallel` will be reserved for other architecture options - parallel pods etc.)
2. Name the env. vars. with `JINA_SHARDS_COMPONENT` where `COMPONENT` is the specific component for which you're setting that number of shards.

JoanFM (approved):

fashion-example-query/README.md
```diff
@@ -43,31 +43,19 @@ python app.py index
...
-
-We index the data updating updating the label
+We will create a separate index per category, so each category will have its own yml file and then we index the data updating updating the label
 
 ```
 d.tags.update({'label': get_mapped_label(label_int)})
 ```
 This is where you could tweak the code if you would like to see only one category.
-For example if you would like to see only dresses, you could do something like this
+For example if you would like to see only Pullovers, you could do something like this
```
cristianmtr, line 58:
```suggestion
For example if you would like to see only pullovers, you could do something like this
```

fashion-example-query/README.md
```diff
@@ -84,21 +72,21 @@ python app.py query
 
 ## Results
 
-This is an example of the results after the query using all categories
+So if we run this as it is, we will see our results filtered by Pullovers like this:
```
cristianmtr, line 75:
```suggestion
So if we run this as it is, we will see our results filtered by pullovers like this:
```

fashion-example-query/README.md
cristianmtr: @theUnkownName

fashion-example-query/README.md
theUnkownName: Done

cristianmtr (approved):

PR #357, opened 2021-01-18
feat: add jinahub image for the advanced vector search example

JoanFM:


florian-hoenicke (approved):
I tested it locally and it works fine.

PR #364, opened 2021-01-22
test: add fashion integration test

Yongxuanzhang:
This pr intends to add an integration test for fashion example and include in ci.

### DIFF
--- a/fashion-example-query/tests/test_fashion_example_search.py
+++ b/fashion-example-query/tests/test_fashion_example_search.py
@@ -1,0 +1,182 @@
+__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__license__ = "Apache-2.0"
+
+import os
+
+import urllib.request
+import gzip
+import numpy as np
+
+from jina.flow import Flow
+from jina.logging.profile import ProgressBar
+from jina import Document
+from jina.executors.encoders import BaseImageEncoder
+
+from pkg_resources import resource_filename
+
+
+class MyEncoder(BaseImageEncoder):
+    """Simple Encoder used in :command:`jina hello-world`,
+        it transforms the original 784-dim vector into a 64-dim vector using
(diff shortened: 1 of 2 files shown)


fashion-example-query/tests/test_fashion_example_search.py
```diff
@@ -0,0 +1,182 @@
...
+        u, s, vh = np.linalg.svd(H, full_matrices=False)
+        self.oth_mat = u @ vh
+        self.touch()
+
+    def encode(self, data: 'np.ndarray', *args, **kwargs):
+        # reduce dimension to 50 by random orthogonal projection
+        return (data.reshape([-1, 784]) / 255) @ self.oth_mat
+
+
+result = []
```
JoanFM, line 36: please make this local to the `test` function

fashion-example-query/tests/test_fashion_example_search.py
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    config('index')
+    download_data(targets, None)
+    index_documents(NUM_DOCS_INDEX, targets)
+
+    config('query')
+    query(NUM_DOCS_QUERY, targets)
+    result_set = set(result)
+    contain_query = True if 'Pullover' in result_set else False
```
JoanFM, line 178: we should check that every single result is a `Pullover`

Yongxuanzhang: Like using a for loop? I think the test now is equal to the 'for loop check' because I will also check if the size of result_set is 1, which means that all of the results are the same and it is 'Pullover'

fashion-example-query/tests/test_fashion_example_search.py
Yongxuanzhang: Did you mean the 'result'?

JoanFM (approved):

theUnkownName (approved):

PR #367, opened 2021-01-22
refactor: add sharding to cxmodal example

JoanFM:


cross-modal-search/app.py
```diff
@@ -12,9 +12,8 @@
 
 
 def config():
-    os.environ['PARALLEL'] = str(1)
-    os.environ['SHARDS'] = str(1)
-    os.environ['COLOR_CHANNEL_AXIS'] = str(0)
+    os.environ['JINA_PARALLEL'] = os.environ.get('JINA_PARALLEL', str(1))
```
cristianmtr, line 15: Why are they both here? There should be only one, right? Also, let's stick to just using the `shards` naming

cristianmtr (changes requested):
naming shards/parallel

cross-modal-search/app.py
JoanFM: to differentiate between encoder and indexer,  I keep shard as parameter of the Pods, but kept the env variable. It can be updated in a next iteration? I have uploaded a docker image with this code and would love to have it in sync

cristianmtr (approved):
sure

PR #370, opened 2021-01-26
fix: update tumblr example

cristianmtr:


tumblr-gif-search/app.py
```diff
@@ -7,16 +7,19 @@
 
 from jina.flow import Flow
 
-# num_docs = os.environ.get('MAX_DOCS', 30)
-
 GIF_BLOB = 'data/*.gif'
+LEN_DOCS = len(glob(GIF_BLOB))
+# allows for returning all docs
 JINA_TOPK = 5000
```
JoanFM, line 13: this will need to be changes when we index a good amount of docs

tumblr-gif-search/flow-index.yml
```diff
@@ -21,8 +21,7 @@ pods:
     separated_workspace: true
```
JoanFM, line 21: remember separatrd workspace is deprcated

tumblr-gif-search/index/_merge_matches_topk.yml
```diff
@@ -0,0 +1,17 @@
...
+with: {}
+metas:
+  name: merge_matches_topk
+requests:
+  on:
+    [SearchRequest, TrainRequest, IndexRequest, DeleteRequest, UpdateRequest]:
+      - !ReduceAllDriver
+        with:
+          traversal_paths: ['m']
+      - !SliceQL
```
JoanFM, line 11: u need to add a SortQL between Reduce and SliceQL

JoanFM (changes requested):

tumblr-gif-search/README.md
```diff
@@ -258,7 +258,7 @@ In this example, we set the number of `parallel` to 8 for `chunk_idx`. After ind
  |-doc_indexer.bin
 ```
 
-You can see that the data is splitted into 8 different directories under the given `workspace`, in a more or less uniform way. This is good because otherwise we end up with one big file that is slow to load and to query. With sharding enabled, one can query multiple smaller index in parallel, which gives better efficiency. 
+You can see that the data is splitted into 8 different directories under the given `workspace`, in a more or less uniform way. This is good because otherwise we end up with one big file that is slow to load and to query. With sharding enabled, one can query multiple smaller index in SHARDS_CHUNK_SEG, which gives better efficiency.
```
JoanFM, line 261: here `parallel` is the right word to use

tumblr-gif-search/README.md
```diff
@@ -176,9 +176,9 @@ def input_fn(with_filename=True):
 
 What it does is reading all gif video files under `GIF_GLOB` and sending the binary contents to the Jina gateway one by one.
 
-One natural question is why can't we send file path (i.e. a tiny binary string) to the gateway and parallelize the file reading inside Jina with multiple crafters?
+One natural question is why can't we send file path (i.e. a tiny binary string) to the gateway and SHARDS_CHUNK_SEGize the file reading inside Jina with multiple crafters?
 
-If your data is stored on HDD, then multiple crafters can not improve the performance: the mechanical structure limits that only one "block" can be read/written at the same time. As your data files probably scatter all over the place, random read/write in parallel won't make significant difference in speed comparing to sequential reading.
+If your data is stored on HDD, then multiple crafters can not improve the performance: the mechanical structure limits that only one "block" can be read/written at the same time. As your data files probably scatter all over the place, random read/write in SHARDS_CHUNK_SEG won't make significant difference in speed comparing to sequential reading.
```
JoanFM, line 181: `parallel` here is the right word

tumblr-gif-search/README.md
```diff
@@ -176,9 +176,9 @@ def input_fn(with_filename=True):
 
 What it does is reading all gif video files under `GIF_GLOB` and sending the binary contents to the Jina gateway one by one.
 
-One natural question is why can't we send file path (i.e. a tiny binary string) to the gateway and parallelize the file reading inside Jina with multiple crafters?
+One natural question is why can't we send file path (i.e. a tiny binary string) to the gateway and SHARDS_CHUNK_SEGize the file reading inside Jina with multiple crafters?
```
JoanFM, line 179: watch with the bulk substitutions

JoanFM (changes requested):

tumblr-gif-search/flow-index.yml
cristianmtr: Removed

tumblr-gif-search/app.py
cristianmtr: From what I've seen in multires, there's no attempt to limit to TOPK. It does seem like it's not really working with the chunking of the gifs. Should we just remove any mention of it from the example?

tumblr-gif-search/README.md
cristianmtr: I've reverted some of these. Can you have a look again?

tumblr-gif-search/index/_merge_matches_topk.yml
```diff
@@ -10,13 +10,13 @@ requests:
...
         with:
-          reverse: False
+          reverse: false
           traversal_paths: ['m']
           field: 'score__value'
       - !SliceQL
         with:
           start: 0
-          end: $JINA_TOPK
+          end: 10
```
theUnkownName, line 19: Why we hardcode this instead of using the topk?

JoanFM: because is updated by the `QuerySetReader` so is missleading

JoanFM (approved):
Do we plan to add `tests` and a docker image with a big dataset indexed?

cristianmtr:
> Do we plan to add `tests` and a docker image with a big dataset indexed?

Not sure. Should all examples have these?

PR #371, opened 2021-01-26
test: add pokedex test

theUnkownName:


pokedex-with-bit/tests/test_pokedex.py
```diff
@@ -0,0 +1,60 @@
...
+def object_image_paths():
+    return ['tests/test-data/1.png', 'tests/test-data/2.png', 'tests/test-data/3.png']
+
+
+def test_query(tmpdir, object_image_paths):
+    config(tmpdir)
+    index_documents()
+    f = get_flow()
+    with f:
+        for object_image_path in object_image_paths:
```
nan-wang, line 57: We can reuse the enviroment variables here.
```suggestion
        from itertools import chain
        from glob import iglob
        for object_image_path in chain.from_iterable(iglob(p) for p in [os.environ['JINA_DATA']):
```

nan-wang (changes requested):

pokedex-with-bit/tests/test_pokedex.py
```diff
@@ -0,0 +1,60 @@
...
+def get_flow():
+    f = Flow().load_config('flows/query.yml')
+    f.use_rest_gateway()
+    return f
+
+
+@pytest.fixture
+def object_image_paths():
+    return ['tests/test-data/1.png', 'tests/test-data/2.png', 'tests/test-data/3.png']
+
```
nan-wang, line 50:
```suggestion
```

pokedex-with-bit/tests/test_pokedex.py
```diff
@@ -0,0 +1,60 @@
...
+
+def call_api(url, payload=None, headers={'Content-Type': 'application/json'}):
+    import requests
+    return requests.post(url, data=json.dumps(payload), headers=headers).json()
+
+
+def get_results(query, top_k=TOP_K):
+    return call_api(
+        'http://0.0.0.0:45678/api/search',
```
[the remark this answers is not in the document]
theUnkownName: Got it

pokedex-with-bit/tests/test_pokedex.py
```diff
@@ -44,17 +44,13 @@ def get_flow():
...
+def test_query(tmpdir):
     config(tmpdir)
     index_documents()
     f = get_flow()
     with f:
-        for object_image_path in object_image_paths:
+        from glob import iglob
+        for object_image_path in iglob(os.environ['JINA_DATA']):
             output = get_results(object_image_path)
             matches = output['search']['docs'][0]['matches']
```
JoanFM, line 55: can we add a check that for every match `uri` is not empty?

JoanFM (changes requested):

pokedex-with-bit/tests/test_pokedex.py
theUnkownName: Done

JoanFM: can u check if this exists for all of the matches and not only for the first one?

JoanFM (changes requested):

JoanFM (changes requested):
Ok now it needs to be added to the CI right?

pokedex-with-bit/tests/test_pokedex.py
```diff
@@ -20,6 +20,10 @@ def config(tmpdir):
     os.environ['JINA_WORKSPACE'] = str(tmpdir)
 
 
+def sh():
```
JoanFM, line 23: u can add the config and download as part of fixtures from which the test actually depends to better separate the preparation from the actual test

JoanFM (changes requested):

nan-wang (approved):

pokedex-with-bit/tests/test_pokedex.py
JoanFM: maybe a better name? download_model?

JoanFM (changes requested):

JoanFM (dismissed):

.github/workflows/ci.yml
```diff
@@ -27,9 +27,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -r ${{ matrix.path }}/requirements.txt --no-cache-dir
           pip uninstall -y jina
           pip install jina[cicd,test,http,hub] requests --no-cache-dir
+          pip install -r ${{ matrix.path }}/requirements.txt --no-cache-dir
```
JoanFM, line 32: I think the point is to make sure `jina` is installed from latest version to keep track if examples work with new versions of jina

JoanFM (changes requested):

.github/workflows/ci.yml
theUnkownName: How you think would be the good way to go around this then? Because the requirements are being overwritten which causes the conflicts

JoanFM: maybe a clear exception for pokedex with installing tf 2.1.0

theUnkownName: Adding something like this at the end of the other installs?
`pip install -r pokedex-with-bit/requirements.txt --no-cache-dir`

.github/workflows/ci.yml
```diff
@@ -27,9 +26,10 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
+          pip install -r ${{ matrix.path }}/requirements.txt --no-cache-dir
           pip uninstall -y jina
           pip install jina[cicd,test,http,hub] requests --no-cache-dir
-          pip install -r ${{ matrix.path }}/requirements.txt --no-cache-dir
+          pip install -r pokedex-with-bit/requirements.txt --no-cache-dir
```
JoanFM, line 32: no, can u add an if matrix_path == pokedex_with_bit install `tensorflow 2.1` or something like this

JoanFM (changes requested):

.github/workflows/ci.yml
theUnkownName: After discussing with the team we thought that add a CI exception for one test only might be sketchy, so I installed the specific version directly on the test, what you think?

JoanFM: It is okey

nan-wang (approved):
LGTM👍

JoanFM (approved):

PR #383, opened 2021-02-05
fix(wikipedia-sentence-search): rm old version

alexcg1:
Remove deprecated version of wikipedia search which uses Jina 0.8.2


nan-wang (approved):
LGTM👍

PR #388, opened 2021-02-09
fix: update to latest jina and runnable

maximilianwerk:
multires-lyrics works with 1.0.0 due to a [recent fix in jina](https://github.com/jina-ai/jina/pull/1917)


alexcg1 (approved):
Woohoo! :+1:

PR #390, opened 2021-02-10
feat: bumped audio version

maximilianwerk:
Worked locally, even thought audio files in jinabox are .ogg which does not work with the audio example. (but that might also be true for 0.9.23)


maximilianwerk:
I'll do another PR with all requirements.txt files updated to 1.0

PR #415, opened 2021-03-04
fix: fix tumblr gif search example

nan-wang:
depends on the next patch release of jina `v1.0.9`, which will include https://github.com/jina-ai/jina-hub/pull/4775


tumblr-gif-search/tests/test_tumblr_integration.py
```diff
@@ -25,40 +25,32 @@ def env_setup(tmpdir):
...
-    return call_api(
-        f'http://0.0.0.0:{os.environ["JINA_PORT"]}/api/search',
-        payload={'top_k': top_k, 'mode': 'search', 'data': [query]})
+        f.index_files(GIF_BLOB, request_size=REQ_SIZE, read_mode='rb')
 
 
 def get_flow():
     f = Flow().load_config('flow-query.yml')
-    f.use_rest_gateway()
+    f.use_grpc_gateway()
```
Yongxuanzhang, line 40: Any reason we use grpc instead of rest here? Can we test both?

Yongxuanzhang (approved):
LGTM 👍

Yongxuanzhang:
The test is much clearer! Should we use use_grpc_gateway or rest for tests?

nan-wang:
> The test is much clearer! Should we use use_grpc_gateway or rest for tests?

Yes, let's do that.

PR #422, opened 2021-03-09
docs(wikipedia-sentences): table format, emoji, gif

alexcg1:
- Fix formatting in table
- Add emojis
- Add gif

Will later share this with folks from Wikimedia, so want it to look good!


wikipedia-sentences/README.md
```diff
@@ -14,14 +16,14 @@
       Output
     </td>
     <td>
-      <pre>top_k</pre> number of sentences that match input query
+      <i>top_k</i> number of sentences that match input query
```
Yongxuanzhang, line 19: We can use markdown table here maybe? Like using this https://www.tablesgenerator.com/markdown_tables

nan-wang (approved):
LGTM👍

PR #423, opened 2021-03-11
simple nlp template to "replace" cookiecutter

alexcg1:
[My First Jina App](https://docs.jina.ai/chapters/my_first_jina_app.html) won't work since we're between deprecating cookiecutter and moving to something better. We still need that doc though. Current plan is to work from a simple template to build a user's first app.

This is that template. Instructions for actually building the app will go into the My First Jina App doc, and not covered here (since if someone writes say, a "how to write a Chinese search app" in Jina they can use this same template and get user to modify)

- feat(template-nlp): initial commit
- feat(template-nlp): initial commit


alexcg1:
@nan-wang @JoanFM could someone review this please? Our [My First Jina App](https://docs.jina.ai/chapters/my_first_jina_app/) is currently broken (due to cookiecutter deprecation) and I urgently want to fix this issue

alexcg1:
Note: This **isn't** a long-term fix. But we don't have a working cookiecutter replacement atm so this slaps a band-aid on the problem and least it works

Yongxuanzhang (approved):
LGTM. So this is an abstract example of the Wikipedia search?

cristianmtr (approved):

alexcg1:
Ugh, our poor CI system :cry:

alexcg1:
> LGTM. So this is an abstract example of the Wikipedia search?

Yup @Yongxuanzhang . Basically a fill-in-the-blanks

Issue #429, opened 2021-03-16
fix examples for Jina==1.1


Yongxuanzhang:
- [x] multires-lyrics-search: this is because [#2141](https://github.com/jina-ai/jina/pull/2141) introduced breaking changes. will be fixed in ->https://github.com/jina-ai/jina-hub/pull/5789
- [x] tumblr-gif-search: same reason as [#2141](https://github.com/jina-ai/jina/pull/2141) -> https://github.com/jina-ai/examples/pull/431 https://github.com/jina-ai/examples/pull/434
- [x] audio-search: same reason as [#2141](https://github.com/jina-ai/jina/pull/2141)-> https://github.com/jina-ai/examples/pull/430
- [x] object-search is fixed since `single_multi_input` is fixed in Imagereader
- [x] pokedex-search is fixed since `single_multi_input` is fixed in Imagereader

The tumblr is failing in 1.0.12 probably because of https://github.com/jina-ai/jina/pull/2198, or other reasons maybe

PR #430, opened 2021-03-16
fix[audio]: adapt segmenter for changes

Yongxuanzhang:
Issue: https://github.com/jina-ai/examples/issues/429
Reason: https://github.com/jina-ai/jina/pull/2141 
Another solution: https://github.com/jina-ai/jina/pull/2172
Ref: 
https://github.com/jina-ai/jina/issues/2176
https://github.com/jina-ai/jina/pull/2178

Need to wait for the changes of batching are done.

### DIFF
--- a/audio-search/pods/customized_executors.py
+++ b/audio-search/pods/customized_executors.py
@@ -5,7 +5,7 @@ import io
 from typing import Any, Dict, List
 
 import numpy as np
-from jina.executors.decorators import batching
+from jina.executors.decorators import batching, single_multi_input
 from jina.executors.encoders.frameworks import BaseTFEncoder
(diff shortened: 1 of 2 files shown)


Yongxuanzhang:
Simple adding @single for segment won't fix this audio example, what should be the correct way to fix it instead of the code that I write now?
@JoanFM @nan-wang

JoanFM:
It is possible that it will break, I am coming with a fix that should make the addition of single good enough

Yongxuanzhang:
> It is possible that it will break, I am coming with a fix that should make the addition of single good enough

Cool, so I will wait for your fix? After that @single should be able to work?

nan-wang (approved):
LGTM👍

PR #436, opened 2021-03-19
fix[tumblr]: change ranker

Yongxuanzhang:
tumblr is failing because we need to adapt bimatch ranker to change 'length' to 'weight' or use other ways of implementation.
This is a PR to demo a workaround for this example.(replace the ranker).
But I prefer to use different rankers in our examples.

### DIFF
--- a/tumblr-gif-search/index/rank.yml
+++ b/tumblr-gif-search/index/rank.yml
@@ -1,4 +1,7 @@
-!BiMatchRanker
+!SimpleAggregateRanker
+with:
+  aggregate_function: 'min'
+  inverse_score: true
 requests:
   on:
     [SearchRequest, IndexRequest]:


rutujasurve94:
Can we merge this if `SimpleAggregateRanker` is the right weighted ranker to use?

Yongxuanzhang:
> Can we merge this if `SimpleAggregateRanker` is the right weighted ranker to use?

For me I just want show that simpleaggregrate ranker can also work if someone want to use this example at current version.

I would prefer to fix the removed length attribute for the bimatchranker

rutujasurve94 (approved):

rutujasurve94:
@Yongxuanzhang sure, we can have a separate PR for fixing the removed length attribute for the `bimatchranker`.

PR #454, opened 2021-04-16
feat: ner search example

cristianmtr:
Idea is to showcase how you can employ NER within neural search.

Example change in results from test, from query being `restaurants in San Francisco`

```
    assert [m[0] for m in ner_results] == [
        'San Francisco cheap food\n',  # more important because matches both value and type of entity
        'Los Angeles cuisine\n',  # more imp. because it matches entity type
        'take-away in my city\n',
    ]
    print('basic results:')
    print(f'### {basic_results}')
    assert [m[0] for m in basic_results] == [
        'take-away in my city\n',
        'San Francisco cheap food\n',
        'Los Angeles cuisine\n',
```

### DIFF
--- a/ner-search/README.md
+++ b/ner-search/README.md
@@ -1,0 +1,31 @@
+query:
+
+restaurants in San Francisco
+
+NER: [San Francisco] --> GEO
+
+matches:
+
+food in [San Francisco] -- higher match because same entity 
+
+restaurants in [Los Angeles] -- higher because same entity type
+
+bars in [New York] -- same as above
+ 
+tomatoes are red -- irrelevant
+ 
+horses on the field -- irrelevant
+
+--- 
+
+inedxing as normal
+
+---
+
+query
+
+we get matches into the ranker
+
+the ranker needs to know the original query -- in order to extract it and get its type
+
+the ranker needs the model name -- passed as argument
--- a/ner-search/manifest.yml
+++ b/ner-search/manifest.yml
@@ -1,0 +1,15 @@
+manifest_version: 1
+name: ner-search
+description: 'Example Jina app for ranking results using a NER model'
+author: Jina AI Dev. Team
+url: https://github.com/jina-ai/examples/tree/master/ner-search
+vendor: Jina AI Limited
+documentation: https://github.com/jina-ai/examples/tree/master/ner-search/wikipedia-sentences
+version: 0.1
+license: apache-2.0
+keywords: [NLP, text, spacy, example, NER]
+type: app
+kind: example
+avatar: None
+platform: "linux/amd64"
+update: "None"
--- a/ner-search/ner_ranker.py
+++ b/ner-search/ner_ranker.py
@@ -1,0 +1,100 @@
+from collections import namedtuple
+from enum import Enum
+from typing import Dict, List, Optional
+
+import spacy
+from jina.executors.rankers import Match2DocRanker
+
+NEREntity = namedtuple('NEREntity', ('value', 'type'))
+
+
+class EntityMatchEnum(Enum):
+    NONE = 0  # no match
+    TYPE = 1  # matches on same type but diff value
+    FULL = 2  # same type and same value
+    FULL_ALL = 3  # all of them match
+
+
+class NERSpacyRanker(Match2DocRanker):
+    # TODO docstring
+    def __init__(
+        self,
+        model='en_core_web_lg',
+        factor_type_match=1.1,
+        factor_full_match=1.4,
+        *args,
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.model = model
+        self.factor_full_match = factor_full_match
+        self.factor_type_match = factor_type_match
+
+    def post_init(self):
+        """Init the Spacy model"""
+        super(NERSpacyRanker, self).post_init()
+        try:
+            self.nlp = spacy.load(self.model)
+        except Exception as e:
+            self.logger.error(e)
+            self.logger.error(
+                f'{self.model} was not found. Please make sure you either install it or train it and provide it yourself'
+            )
+
+    def score(
+        self,
+        old_matches_scores: List[List[float]],
+        queries_metas: List[Dict],
+        matches_metas: List[List[Dict]],
+    ) -> List[List[float]]:
+        """Calculate the negative Levenshtein distance
+
+        :param old_matches_scores: Contains old scores in a list
+        :param queries_metas: Dictionary containing all the query meta information requested by the `query_required_keys` class_variable.
+        :param matches_metas: List containing all the matches meta information requested by the `match_required_keys` class_variable. Sorted in the same way as `old_match_scores`
+        :return: An iterable of the new scores
+        """
+        return [
+            self._recalc_score_matches(matches_metas[i], old_matches_scores[i], query_meta)
+            for i, query_meta in enumerate(queries_metas)
+        ]
+
+    def _recalc_score_matches(self, matches_metas, old_matches_scores, query_meta):
+        query_ents = [NEREntity(value=e.text, type=e.label_) for e in self.nlp(query_meta['text'].strip()).ents]
+        new_scores = []
+        # TODO design formula for recalculating score
+
+        for match_score, match in zip(old_matches_scores, matches_metas):
+            match_ents = [NEREntity(value=e.text, type=e.label_) for e in self.nlp(match['text'].strip()).ents]
+            entity_match_type = self._entity_match(match_ents, query_ents)
+
+            new_match_score = self._recalc_match_score(match_score, entity_match_type)
+            print(
+                f'#### match text = {match["text"].strip()}; orig score = {match_score}; new score = {new_match_score}; match type = {entity_match_type}'
+            )
+            new_scores.append(new_match_score)
+
+        return new_scores
+
+    def _entity_match(self, match_ents: List[NEREntity], query_ents: List[NEREntity]) -> int:
+        ner_matches = []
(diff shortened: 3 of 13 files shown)


ner-search/ner_ranker.py
```diff
@@ -0,0 +1,24 @@
...
+from jina.executors.rankers import BaseRanker
+
+
+class NERSpacyRanker(BaseRanker):
+    # TODO docstring
+    def __init__(
+        self,
+        model='en_core_web_lg',
+        fallback_type=True,
+        query_required_keys: Optional[List[str]] = None,
```
JoanFM, line 12: no need to add these kwargs if theu exist in the parent

ner-search/pods/ner_ranker.yml
```diff
@@ -0,0 +1,14 @@
...
+with:
+  model: en_core_web_lg
+  fallback_type: True
+  query_required_keys: ['text']
+  match_required_keys: ['text']
+metas:
+  name: ranker
+on:
+  SearchRequest:
+    - !BaseRankDriver
```
JoanFM, line 11: Matches2DocRankDriver

JoanFM (commented):
I like the idea

ner-search/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+NER: [San Francisco] --> GEO
+
+matches:
+
+food in [San Francisco] -- higher match because same entity 
+
+restaurants in [Los Angeles] -- higher because same entity type
+
+bars in [New York] -- same as above
```
maximilianwerk, line 13: This is arguable. You could also say: Should be removed, since wrong entity. Someone searching for restaurants in San Francisco most probably don't want to see any restaurant in their result set in New York.

ner-search/requirements.txt
```diff
@@ -0,0 +1,4 @@
+click==7.1.2
+transformers==4.1.1
+torch==1.7.1
+jina[scipy, http]==1.1.0
```
maximilianwerk, line 4: spacy should be somewhere

ner-search/manifest.yml
```diff
@@ -0,0 +1,15 @@
+manifest_version: 1
+name: wikipedia-sentences-30k
+description: 'Example Jina app for searching 30,000 sentences from Wikipedia'
+author: Alex Cureton-Griffiths (alex.cg@jina.ai)
+url: https://github.com/jina-ai/examples/tree/master/wikipedia-sentences
+vendor: Jina AI Limited
+documentation: https://github.com/jina-ai/examples/tree/master/wikipedia-sentences
+version: 0.2.10
+license: apache-2.0
+keywords: [NLP, wikipedia, text, distilbert, example, transformers]
```
maximilianwerk, line 10: Should be changed.

ner-search/ner_ranker.py
```diff
@@ -0,0 +1,95 @@
...
+        :param matches_metas: List containing all the matches meta information requested by the `match_required_keys` class_variable. Sorted in the same way as `old_match_scores`
+        :return: An iterable of the new scores
+        """
+        return [
+            self._recalc_score_matches(matches_metas[i], old_matches_scores[i], query_meta)
+            for i, query_meta in enumerate(queries_metas)
+        ]
+
+    def _recalc_score_matches(self, matches_metas, old_matches_scores, query_meta):
+        query_ents = [NEREntity(value=e.text, type=e.label_) for e in self.nlp(query_meta['text'].strip()).ents]
```
maximilianwerk, line 58: should be a parameter to the `_recalc_score_matches` function and only calculated once.

ner-search/ner_ranker.py
```diff
@@ -0,0 +1,95 @@
+from collections import namedtuple
+from enum import Enum
+from typing import Dict, List, Optional
+
+import spacy
+from jina.executors.rankers import Match2DocRanker
+
+NEREntity = namedtuple('NEREntity', 'value type')
```
maximilianwerk, line 8: It's a matter of taste, but I find `NEREntity = namedtuple('NEREntity', ('value', 'type'))` better readable

ner-search/ner_ranker.py
```diff
@@ -0,0 +1,95 @@
...
+    ):
+        super().__init__(*args, **kwargs)
+        self.model = model
+        self.factor_full_match = factor_full_match
+        self.factor_type_match = factor_type_match
+
+    def post_init(self):
+        """Init the Spacy model"""
+        super(NERSpacyRanker, self).post_init()
+        # TODO check if isntalled, if not downlaod
```
maximilianwerk, line 36: typo. And: "if not: raise Exception"

ner-search/ner_ranker.py
```diff
@@ -0,0 +1,95 @@
...
+        ]
+
+    def _recalc_score_matches(self, matches_metas, old_matches_scores, query_meta):
+        query_ents = [NEREntity(value=e.text, type=e.label_) for e in self.nlp(query_meta['text'].strip()).ents]
+        new_scores = []
+        # TODO design formula for recalculating score
+
+        for match_score, match in zip(old_matches_scores, matches_metas):
+            match_ents = [NEREntity(value=e.text, type=e.label_) for e in self.nlp(match['text'].strip()).ents]
+            entity_match_score = self._entity_match(match_ents, query_ents)
```
maximilianwerk, line 64: I'd rename it `entity_match_type`. It is not a score, but a category.

maximilianwerk (commented):
I have mixed feelings about it and would love to see this in action (and finetuned) for one of our real-world examples. Following reasons:

- finetuning `factor_type_match` and `factor_full_match` must be done with a lot of care and with each model, you can possibly use.
- I wonder if a hard filtering (e.g. having additive changes to the score instead of multiplicative) would be easier to grasp for the user.

cristianmtr:
> I have mixed feelings about it and would love to see this in action (and finetuned) for one of our real-world examples. Following reasons:
> 
> * finetuning `factor_type_match` and `factor_full_match` must be done with a lot of care and with each model, you can possibly use.

Yes, that's why they are parameters. You could use the Optimizer for this, right?

> * I wonder if a hard filtering (e.g. having additive changes to the score instead of multiplicative) would be easier to grasp for the user.

Could be another argument we can allow to the Ranker

ner-search/README.md
cristianmtr: I think this could be achieved if you penalize `factor_type_match`

ner-search/ner_ranker.py
cristianmtr: It is calculated once per query

ner-search/ner_ranker.py
cristianmtr: done

ner-search/ner_ranker.py
cristianmtr: good idea, thanks

ner-search/ner_ranker.py
cristianmtr: ah yes, good point

ner-search/manifest.yml
cristianmtr: yes, was just copy-paste

ner-search/requirements.txt
cristianmtr: good point

ner-search/ner_ranker.py
maximilianwerk: This is actually only true for `traversal_paths=['r']`. In the moment you choose `traversal_paths=['c']` it is not true anymore. Anyhow, this is a problem of the current Executor/Driver design, so just ignore my comment.

maximilianwerk:
> > I have mixed feelings about it and would love to see this in action (and finetuned) for one of our real-world examples. Following reasons:
> > 
> > * finetuning `factor_type_match` and `factor_full_match` must be done with a lot of care and with each model, you can possibly use.
> 
> Yes, that's why they are parameters. You could use the Optimizer for this, right?

Yes, in principle that is true. We need to check how well that works, especially with changing a model at the same time.

> 
> > * I wonder if a hard filtering (e.g. having additive changes to the score instead of multiplicative) would be easier to grasp for the user.
> 
> Could be another argument we can allow to the Ranker

:+1:

maximilianwerk:
Some more thoughts:

1. Please also commit the toy-data file.
2. Since this is "testing a new idea" and not an "user can learn from it directly" example, I'd suggest we move it into another repository, until we are certain, that we get quality results on a proper testset. From my understanding the example repo is for ready-to-use examples for our users.
3. Wherever we move this tryout example, I'd like to have it run on a bigger dataset including (!) an evaluation set. When we add such a feature, we do in order to improve the quality. For improving the quality, we need to measure it.
4. Once it is ready-to-use, the ranker should go into the hub. The rest of the code can go into examples and we can demonstrate via the evaluation data, how much quality gain we get over just normal ranking.

FionnD:
> 2. Since this is "testing a new idea" and not an "user can learn from it directly" example, I'd suggest we move it into another repository, until we are certain, that we get quality results on a proper testset. From my understanding the example repo is for ready-to-use examples for our users.

Yes please create your own repo for this for the moment. Once it reaches a level of quality that we are happy with then we move it here

cristianmtr:
> > 1. Since this is "testing a new idea" and not an "user can learn from it directly" example, I'd suggest we move it into another repository, until we are certain, that we get quality results on a proper testset. From my understanding the example repo is for ready-to-use examples for our users.
> 
> Yes please create your own repo for this for the moment. Once it reaches a level of quality that we are happy with then we move it here

Sure thing

cristianmtr:
Moved to https://github.com/jina-ai/ner-search

Issue #457, opened 2021-04-20
add QPS timing scripts to each example


cristianmtr:
Follow steps defined in https://github.com/jina-ai/examples/pull/449

- add `TimeContext` to index and query ops:

```python
            with TimeContext(f'QPS: indexing {len(list(data_func_list))}', logger=flow.logger):
```

and

```python
            with TimeContext(f'QPS: query with {len(query_input)}', logger=flow.logger):
```

- add `setup_run.sh` script to example directory that prepares and runs the example, while piping the output into a `metrics.txt` file

ex.:

```bash
#!/bin/bash
# these scripts will contain everything that needs to be run before the example
# downloading data, model, preprocessing etc.
bash ./get_data.sh siftsmall && \
rm -rf workspace && \
bash ./generate_training_data.sh && \
python app.py -t index | tee metrics.txt && \
python app.py -t query | tee >> metrics.txt
```

Create a PR and edit this comment to add next to the example

- [x] advanced-vector-search https://github.com/jina-ai/examples/pull/449
- [x] audio-search https://github.com/jina-ai/examples/pull/449
- [x] chinese-text-search https://github.com/jina-ai/examples/pull/459
- [x] cross-modal-search https://github.com/jina-ai/examples/pull/459
- [x] fashion-example-query https://github.com/jina-ai/examples/pull/459
- [x] multimodal-search-pdf
- [x] multimodal-search-tirg https://github.com/jina-ai/examples/pull/459
- [x] multires-lyrics-search https://github.com/jina-ai/examples/pull/459
- [ ] ner-search
- [x] object-search https://github.com/jina-ai/examples/pull/477
- [x] pokedex-with-bit https://github.com/jina-ai/examples/pull/460
- [ ] southpark-search
- [x] tumblr-gif-search https://github.com/jina-ai/examples/pull/460
- [x] wikipedia-sentences https://github.com/jina-ai/examples/pull/460
- [x] wikipedia-sentences-incremental https://github.com/jina-ai/examples/pull/460

slettner:
@cristianmtr The `ner-search` and `southpark-search` examples are not present anymore, correct?
If yes, I think this issue is done.

cristianmtr:
> @cristianmtr The `ner-search` and `southpark-search` examples are not present anymore, correct?
> If yes, I think this issue is done.

Yes, those were remnants of previous branches in my local repo. This is done! Good job, everyone :100:

PR #459, opened 2021-04-21
feat: timing examples

jacobowitz:
This adds QPS timing for a couple of examples as defined in https://github.com/jina-ai/examples/issues/457

The examples are:

- chinese-text-search
- cross-modal-search
- fashion-example-query
- multimodal-search-tirg
- multires-lyrics-search

There is no timing support for querying for cross-modal-search and multires-lyrics-search because they are blocking.

Two datasets were relying on kaggle, which does not really work in this automated case, I copied those datasets to google drive. I dont really like that, but I did not see another easy solution.

Besides the timing, I also fixed some bug in multimodal-search-tirg/flow-query.yml definition. Please double check that change. For me the reference to the yaml file was broken before as folder yaml/ does not exist.


JoanFM:
Hey @FionnD , @cristianmtr @maximilianwerk. Would it be cool to show how we can extrac statistics about the `time` spent in every `Pod`? Like this we can start gaining more useful information on potential optimizations and chanegs?

chinese-text-search/app.py
```diff
@@ -35,21 +35,23 @@ def index(num_docs):
...
 
 
 def query(top_k):
     f = Flow().load_config("flows/query.yml")
     with f:
-        while True:
-            text = input("please type a sentence: ")
-            if not text:
-                break
+        with TimeContext(f'QPS: query', logger=f.logger):
```
cristianmtr, line 45: No need for TimeContext when it's blocking

chinese-text-search/setup_run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+rm -rf workspace && \
+python app.py -t index | tee metrics.txt && \
+python app.py -t query <<< 'some text from stdin' | tee >> metrics.txt
```
cristianmtr, line 4: oh I see. You're simulating sending a query via stdin. Then ignore the above comment, if this works well

jacobowitz: yeah I think it should work, at least metrics.txt looked okay. The app should also exit after reading the string so there should be no blocking here

jacobowitz:
Tests are failing due to this
`Module load error jina.hub.indexers.vector.PysparnnIndexer No module named 'pysparnn'` which looks unrelated to this change?

Yongxuanzhang:
> Tests are failing due to this
> `Module load error jina.hub.indexers.vector.PysparnnIndexer No module named 'pysparnn'` which looks unrelated to this change?

Not this one, we need to update all examples to adapt to changes in the latest release

Yongxuanzhang:
> Tests are failing due to this
> `Module load error jina.hub.indexers.vector.PysparnnIndexer No module named 'pysparnn'` which looks unrelated to this change?

multimodal-search-tirg/app.py
```diff
@@ -94,11 +95,13 @@ def main(task, data_path, num_docs, batch_size, image_path, text_query, overwrit
...
         f = Flow.load_config('flow-index.yml')
         with f:
-            f.index(index_generator(data_path, num_docs), batch_size=batch_size)
+            with TimeContext(f'QPS: indexing {num_docs}', logger=f.logger):
+                f.index(index_generator(data_path, num_docs), batch_size=batch_size)
     elif task == 'query':
         f = Flow.load_config('flow-query.yml')
         with f:
-            f.search(input_fn=query_generator(image_paths, text_queries), on_done=print_result)
+            with TimeContext(f'QPS: query', logger=f.logger):
```
cristianmtr, line 103: If we keep the QPS tracking on the blocking request we need to specify that it's for only 1 document, while here it's for `len(image_paths)`

multires-lyrics-search/app.py
```diff
@@ -42,7 +43,8 @@ def input_fn():
 def index():
     f = Flow.load_config('flows/index.yml')
     with f:
-        f.index(input_fn, request_size=8)
+        with TimeContext(f'QPS: indexing', logger=f.logger):
```
cristianmtr, line 46: same, specify the nr of docs

cristianmtr (changes requested):
some comments + to find out whether we can host the datasets

cristianmtr (changes requested):
some comments + to find out whether we can host the datasets

multires-lyrics-search/app.py
jacobowitz: hmm but I dont know it upfront in this case do I? We use the generator `input_fn()  ` based on some csv file.

multimodal-search-tirg/app.py
jacobowitz: Sure will do here.

For the blocking case with the single document input from stdin (chinese text search), I cant really know though that it is a single document? In our context of course we know, but in code the time tracking is started outside reading from stdin so a user can feed it in multiple documents?

multires-lyrics-search/app.py
cristianmtr: You should look at that file and count the lines maybe. Or make the generator into a list and get its `len()`. (yes, not efficient, but needed in this case)

multimodal-search-tirg/app.py
cristianmtr: I think you can assume it's only one, right? Or at least in our case (in the `.sh` script) it's only 1 doc, since it's done with this call

```
                f.search_lines(lines=[text, ], line_format='text', on_done=ppr, top_k=top_k)
```

nan-wang (approved):
LGTM👍

slettner:
S3 code for downloading data-sets coming in https://github.com/jina-ai/examples/pull/538

jacobowitz:
I am using @slettner code now for S3 access for downloading datasets instead of google drive. 
Also rebased this PR against current master

jacobowitz:
@cristianmtr 
can you have another look please? Rebase/S3 usage was added since your last review. 

We should check that all examples run though once datasets are all uploaded. Will do that with @slettner

cross-modal-search/app.py
```diff
@@ -87,10 +88,11 @@ def main(task, num_docs, request_size, data_set, model_name):
     print(f'### task = {task}')
```
cristianmtr, line 88: not sure why the previous PR was merged with these prints, but can you remove them?

cristianmtr (dismissed):
nice!

cross-modal-search/app.py
jacobowitz: I can, but it looks like they are all over the place? Should we remove all, maybe lets configure a logger and replace the prints with logging instead?

cristianmtr: Depends on what you're printing. They would also need to be consistent across examples, I'd expect. But yes, go ahead, from my part. Do remove the printing of the top k I saw somewhere, as it pollutes the CI check

chinese-text-search/app.py
```diff
@@ -68,7 +69,8 @@ def query(top_k):
             def ppr(x):
                 print_topk(x, text)
```
cristianmtr, line 70: Here

multires-lyrics-search/app.py
```diff
@@ -41,14 +44,18 @@ def input_fn():
                     d.tags['SName'] = row[1]
                     d.tags['SLink'] = row[2]
                     d.text = row[3]
-                yield d
+                docs.append(d)
+
+    return docs
```
Yongxuanzhang, line 49: any reason we want to use list instead of generator?

jacobowitz: I think because its neat to get the total size upfront for the logging, but I guess its not really necessary.

Yongxuanzhang: oh ok

Yongxuanzhang (dismissed):

jacobowitz:
I rebased and ran the examples this Pr touches manually:


-  [x] chinese text search
-  [x] cross modal search
-  [] fashion example query (fails, but I think unrelated)
- [] multimodal search tirg (still running)
- [] multires lyric search (fails, probably this PRs fault)

Yongxuanzhang (approved):

PR #463, opened 2021-04-22
fix: chinese text eg using CRUD REST

rutujasurve94:
- update Jina to 1.1.0 (if not already at that version or higher)

- modify flow yamls to enable REST API

- add index_rest method that indexes docs using the /index REST API

- add index_rest as CLI option to -t (via click options)

- change README to point towards /search (NOT /api/search)


nan-wang (approved):
LGTM👍

PR #464, opened 2021-04-22
fix: audio search eg using CRUD REST

rutujasurve94:
update Jina to 1.1.0 (if not already at that version or higher)

modify flow yamls to enable REST API

add index_rest method that indexes docs using the /index REST API

add index_rest as CLI option to -t (via click options)

change README to point towards /search (NOT /api/search)

### DIFF
--- a/audio-search/app.py
+++ b/audio-search/app.py
@@ -1,4 +1,4 @@
-__copyright__ = "Copyright (c) 2020 Jina AI Limited. All rights reserved."
+__copyright__ = "Copyright (c) 2021 Jina AI Limited. All rights reserved."
 __license__ = "Apache-2.0"
 
 __version__ = '0.0.1'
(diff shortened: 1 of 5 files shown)


audio-search/app.py
```diff
@@ -17,14 +20,34 @@
...
+        url = f'http://0.0.0.0:{f.port_expose}/index'
+
+        input_docs = _input_lines(
+            filepath=data_path,
+            size=num_docs,
+            read_mode='r',
+        )
+        data_json = {'data': [Document(text=text).dict() for text in input_docs]}
+        print(f'#### {len(data_json["data"])}')
+        with TimeContext(f'QPS: indexing {num_docs}', logger=f.logger):
```
slettner, line 46: I could happen that `num_docs` is larger than the actual number of available docs. For example, when the user enters a very high number. We could take the min here to make sure that the QPS print is correct.

```
min(num_docs, len(data['data'])
```

rutujasurve94: this is an edge case, as the user knows the dataset for indexing before hand, like say `flickr30k`, user shouldn't put `num_docs` as `40000`. 
however, this could be considered for refactoring all examples for consistency.

cristianmtr (approved):

PR #477, opened 2021-04-26
Feat timing examples 3

slettner:
Add performance measure to the `object-search` example as suggested in #457

Note: Waiting on clarification of licensing for the kaggle data-sets to decide download procedure. Currently relying on toy data (See https://github.com/jina-ai/examples/pull/538).


object-search/app.py
```diff
@@ -4,44 +4,42 @@
...
 
 
 def config():
-    parallel = 1 if sys.argv[1] == 'index' else 1
-    shards = 1
-
-    os.environ['PARALLEL'] = str(parallel)
-    os.environ['SHARDS'] = str(shards)
+    os.environ["JINA_DATA_FILE"] = os.environ.get("JINA_DATA_FILE", IMAGE_SRC)
+    os.environ['PARALLEL'] = str(1)
```
jacobowitz, line 23: I guess we dont need the str conversion here anymore? same for below. I know it was there before e.g. for the port, but it looks kinda weird?

```suggestion
    os.environ['PARALLEL'] = '1'
```

object-search/app.py
```diff
@@ -4,44 +4,42 @@
...
+            f.index_files(IMAGE_SRC, request_size=10, read_mode='rb', size=num_docs)
 
 
 def index_restful(num_docs):
     f = Flow().load_config('flow-index.yml')
 
     with f:
         data_path = os.path.join(os.path.dirname(__file__), os.environ.get('JINA_DATA_FILE', None))
-        print(f'Indexing {data_path}')
+        logger.info(f'Indexing {data_path}')
```
jacobowitz, line 42: I am actually not sure here, but maybe its better to use `f.logger` here instead of the default logger? Same for the other logger usages where we have the flow object already. On the other hand the flow is not available yet in main and you need to use some other logger there.

object-search/app.py
```diff
@@ -72,51 +70,35 @@ def dryrun():
 @click.command()
 @click.option(
     '--task',
-    '-task',
     '-t',
-    type=click.Choice(['index', 'query', 'index_restful', 'query_restful', 'dryrun'], case_sensitive=False),
-)
+    type=click.Choice(['index', 'index-restful', 'query-restful', 'dry'], case_sensitive=False))
```
jacobowitz, line 74: If you change this (renaming and especially removing query without rest support), please update the readme of this example as well as it is mentioning those options.

object-search/app.py
```diff
@@ -72,51 +70,35 @@ def dryrun():
...
-    elif task == 'query':
-        f = Flow.load_config(f'flow-query-{return_image}.yml')
-        with f:
-            f.block()
-    elif task == 'query_restful':
-        query_restful(return_image)
-    elif task == 'dryrun':
+    if task == 'query-restful':
+        if not os.path.exists(workspace):
+            logger.error(f"The directory {workspace} does not exist. Please index first via `python app.py -t index`")
```
jacobowitz, line 98: I know that the string was there before, but if you touch the line anyway, can you please use single quote strings as per our standard?

jacobowitz (changes requested):
some minor comments, but please update the readme as well to reflect your changes :)

object-search/app.py
slettner: Thanks, https://github.com/jina-ai/examples/pull/477/commits/563e81154ab69492cab4d5db3d6d24de77922b0d

object-search/app.py
slettner: https://github.com/jina-ai/examples/pull/477/commits/d83338770de9b7fd5282cc2f12fda47cd0381be9

object-search/app.py
slettner: I returned to the underscore to be consistent with the other examples and updated the README
https://github.com/jina-ai/examples/pull/477/commits/16acf207013f5e0b288cb58c9367ad8dd5cb7301

object-search/app.py
slettner: Sure, I changed it in the file: https://github.com/jina-ai/examples/pull/477/commits/fcfb02eccad7272d342a11e50412fe9cf1cf161d

What do you mean with touch the line away?😀

jacobowitz (approved):
lgtm, thx for addressing the comments!

PR #496, opened 2021-04-28
fix: update links for wikipedia

rutujasurve94:
Adding and fixing links for Wikipedia Examples
Removing irrelevant links


alexcg1 (approved):
LGTM :+1:

PR #513, opened 2021-04-29
fix: fix resource path

JoanFM:


FionnD (approved):
This works on my machine.

PR #514, opened 2021-04-29
feat: add kaggle library

FionnD:
As a user, I would expect all the libraries I need to be in the requirements.txt file.

### DIFF
--- a/object-search/requirements.txt
+++ b/object-search/requirements.txt
@@ -2,3 +2,4 @@ jina[http]==1.1.0
 torch==1.6.0
 torchvision==0.7.0
 Pillow==8.1.1
+kaggle==1.5.12


bwanglzu (approved):
lgtm!

alexcg1:
Reason I left it out was because:

a) Bit more work for user - need to set up API key
b) If user only wants to play with `toy-data.txt` then no need for Kaggle at all
c) If user clones examples repo just to build their own thing with own dataset, again, no need for Kaggle

## Workaround

Note: not a suggestion bc it has downsides

- We look into license of the big dataset (I'm sure it'll be fine to use)
- We host it ourself so users don't need Kaggle (hosting could be it's own issue though - it's a big dataset and if we get popular it could cost a fair bit)

FionnD:
Maybe I'm wrong, but for this example, there is no toy data so a user has to use Kaggle right? I agree that if the step is optional, we should streamline the process. 

Self-hosting, yes we should. The engineering effort is not the problem, I am hesitate to do this without legal approval. I suppose I can talk to Bing before.

PR #516, opened 2021-04-29
fix: remove shuf command 

FionnD:
Removed shuf command since it doesn't exist on Mac. 

Alex, please let me know if you think the new logic in the data file is okay!

### DIFF
--- a/wikipedia-sentences/README.md
+++ b/wikipedia-sentences/README.md
@@ -46,6 +46,8 @@ To index the [full dataset](https://www.kaggle.com/mikeortman/wikipedia-sentence
 5. Delete the old index: `rm -rf workspace`
 6. Index your new dataset: `python app.py -t index`
 
+If you are using a subset of the data (less then 30,000 documents) we recommend you shuffle the data to prevent only indexing a specific letter of the alphabet. On linux on you can use the `shuf` command. 
+
 ## 🔍 Search
 
 ### With REST API
@@ -80,7 +82,7 @@ This will create a Docker image with pre-indexed data and an open port for REST 
 
 ### Image name format
 
-Please use the following name format for your Docker image, otherwise it will be rejected if you want to push it to Jina Hub. 
+Please use the following name format for your Docker image, otherwise it will be rejected if you want to push it to Jina Hub.
 
 ```
 jinahub/type.kind.image-name:image-version-jina_version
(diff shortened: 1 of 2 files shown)


wikipedia-sentences/README.md
```diff
@@ -46,6 +46,8 @@ To index the [full dataset](https://www.kaggle.com/mikeortman/wikipedia-sentence
 5. Delete the old index: `rm -rf workspace`
 6. Index your new dataset: `python app.py -t index`
 
+If you are using a subset of the data (less then 30,000 documents) we recommend you shuffle the data to prevent only indexing a specific letter of the alphabet. On linux on you can use the `shuf` command. 
+
```
alexcg1, line 50: ## Rewrite

(Let me know if it's too wordy)

If you are using a subset of the data (less then 30,000 documents) we recommend you shuffle the data. This is because the input file is ordered alphabetically, and Jina indexes from the top down. So without shuffling, your index may contain unrepresentative data, like this:

```
0.000123, which corresponds to a distance of 705 Mly, or 216 Mpc.
000webhost is a free web hosting service, operated by Hostinger.
0010x0010 is a Dutch-born audiovisual artist, currently living in Los Angeles.
0-0-1-3 is an alcohol abuse prevention program developed in 2004 at Francis E. Warren Air Force Base based on research by the National Institute on Alcohol Abuse and Alcoholism regarding binge drinking in college students.
0.01 is the debut studio album of H3llb3nt, released on February 20, 1996 by Fifth Colvmn Records.
001 of 3 February 1997, which was signed between the Government of the Republic of Rwanda, and FAPADER.
003230 is a South Korean food manufacturer.
```

On Linux, you can shuffle using the [`shuf` command](https://linuxhint.com/bash_shuf_command/):

```bash
shuf input.txt > input.txt
```

To shuffle a file on macos, please read [this post](https://apple.stackexchange.com/questions/142860/install-shuf-on-os-x/195387).

alexcg1 (changes requested):

alexcg1:
I think in the medium-long run it would make sense to retire this data file altogether since it's so broad (as I've discussed on Slack).

I've been testing alternatives and will keep everyone posted

wikipedia-sentences/README.md
FionnD: Way better! Accepted and added.

alexcg1 (approved):
LGTM :+1:

PR #517, opened 2021-04-30
fix: obj search ws

cristianmtr:
Fix a bug that was preventing indexing.

Also moved the logic for checking the workspace to before the actual task calls.


FionnD (approved):
I tested this locally and it solves the issue. Thanks @cristianmtr !

The batching is now hanging, but I will consider this a separate issue.

PR #538, opened 2021-05-03
feat: add s3 script

### LINKED ISSUE #537: Add s3 script
Add script that allows us to download data-sets from an internal S3 bucket. 
This is required for the performance testing to download the data automatically when executing the measurement scripts.

Note: To successfully run the script in the CI, credential need to be provided via system env variables (https://github.com/jina-ai/examples/issues/536).

slettner:
Closes #537


pull_dataset.py
```diff
@@ -0,0 +1,38 @@
+# This script is only used by the CI Pipeline
+import logging
+import os
+
+import click
+from jinacld_tools.aws.services.s3 import S3
+
+
+BUCKET_NAME = "kaggle-data-set-bucket"
```
slettner, line 9: @jacobowitz Can you send be the name of the bucket we want to use for storing the data-sets, then I will add it here

jacobowitz: sure, lets use _jina-examples-datasets_

slettner: Thanks, https://github.com/jina-ai/examples/pull/538/commits/598f7cb76f3a7254595e121b8e05ea31af0db814

pull_dataset.py
```diff
@@ -10,6 +10,13 @@
 log = logging.getLogger(__name__)
 
 
+def _check_credentials_exist():
+    assert os.environ.get('AWS_ACCESS_KEY_ID') is not None, 'Missing access key'
```
jacobowitz, line 14: Maybe improve those error messages a bit? Users might see them and we can actually tell them more, like which env variable specifically is not set.
Same goes for the one below.

jacobowitz (changes requested):

pull_dataset.py
slettner: Good point, hope this looks better https://github.com/jina-ai/examples/pull/538/commits/43ab7fb24b2ed12033dc77c74c45d8ed5d8eaacf

jacobowitz (approved):
lgtm :+1: 
Please do squash merge when you merge :)

PR #549, opened 2021-05-06
fix: multires lyrics

rutujasurve94:
- update `app.py`

### DIFF
--- a/multires-lyrics-search/app.py
+++ b/multires-lyrics-search/app.py
@@ -95,25 +95,13 @@ def dryrun():
 @click.command()
 @click.option('--task', '-t',
               type=click.Choice(['index', 'query', 'index_restful', 'query_restful', 'dryrun'], case_sensitive=False))
-@click.option('--num_docs_query', '-n', default=100)
 @click.option('--num_docs_index', '-n', default=600)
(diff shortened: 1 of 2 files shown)


multires-lyrics-search/app.py
```diff
@@ -97,8 +97,8 @@ def dryrun():
               type=click.Choice(['index', 'query', 'index_restful', 'query_restful', 'dryrun'], case_sensitive=False))
 @click.option('--num_docs_query', '-n', default=100)
 @click.option('--num_docs_index', '-n', default=600)
-def main(task, num_docs_query, num_docs_index):
-    config(task)
+def main(task, num_docs_index):
```
jacobowitz, line 100: `num_docs_query` isnt used anymore now is it? If so then lets remove it from the parameters?

jacobowitz (changes requested):
I think we also need to update the readme? It is missing the -t parameter in some descriptions (run section).

Also the the config method tries to do this

`parallel = 2 if sys.argv[1] == 'index' else 1`
I think that does not work anymore with `python app.py -t index ` ?

jacobowitz (approved):
Lgtm. Thx for updating the readme as well :)

PR #550, opened 2021-05-06
fix: cross modal example

rutujasurve94:
- Fix `cross modal` search for `num_docs` and `request size`


jacobowitz (changes requested):
lets use single quotes instead of double quotes for strings? otherwise looks good to me :)

jacobowitz (approved):

PR #552, opened 2021-05-06
refactor: logging and restructure workspace cleaning

cristianmtr:
Refactoring prints to logger

Refactoring structure of code (around workspace-checking)

### DIFF
--- a/object-search/app.py
+++ b/object-search/app.py
@@ -11,11 +11,15 @@ import click
 from jina.flow import Flow
 from jina import Document
 from jina.clients.sugary_io import _input_lines
-from jina.logging import default_logger as logger
 from jina.logging.profile import TimeContext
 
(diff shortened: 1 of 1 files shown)


object-search/app.py
```diff
@@ -62,43 +58,61 @@ def query_restful(return_image):
 
 
 def dryrun():
-    f = Flow().load_config('flow-index.yml')
+    f = Flow().load_config("flow-index.yml")
```
jacobowitz: single quotes

jacobowitz (approved):
lgtm. The auto workspace removal is great.

While we are already doing quality of life improvements, maybe it makes sense to have a joined entrypoint for index and query as well? But maybe we also want to explain those concepts separately and leave both separate?

Yongxuanzhang (approved):

PR #578, opened 2021-05-12
feat: minor improvements

FionnD:
Some minor improvements from other PRs

### DIFF
--- a/example_template.md
+++ b/example_template.md
@@ -24,15 +24,6 @@
 | Model used | *Link to the model* |
 
 
-This example shows you how to...
-
-*Add a short description of*
-1. *What is this example?*
-2. *What is the expected outcome of this example?*
-3. *What the user will learn in this example?*
-
-
-
 ## 🐍 Build the app with Python
 
 These instructions explain how to build the example yourself and deploy it with Python. If you want to skip the building steps and just run the example with Docker, check [the Docker deployment instructions at the end of this README](#deploy-with-docker)  
--- a/example_template.md
+++ b/example_template.md
@@ -3,6 +3,16 @@
 
 
  *ADD A TABLE OF CONTENTS HERE *
+ 
+ - [Overview](#overview)
+- [🐍 Build the app with Python](#-build-the-app-with-python)
+- [🔮 Overview of the files in this example](#-overview-of-the-files-in-this-example)
+- [🌀 Flow diagram](#-flow-diagram)
+- [🔨 Next steps, building your own app](#-next-steps-building-your-own-app)
+- [🐳 Deploy the prebuild application using Docker](#-deploy-the-prebuild-application-using-docker)
+- [🙍 Community](#-community)
+- [🦄 License](#-license)
+
 
 ## Overview
 | About this example: |  |
--- a/example_template.md
+++ b/example_template.md
@@ -7,7 +7,9 @@
 | Dataset used | *Link to the datasets* |
 | Model used | *Link to the model* |
 
-<p></p>
+
+ *ADD A TABLE OF CONTENTS HERE *
+
 This example shows you how to...
 
 *Add a short description of*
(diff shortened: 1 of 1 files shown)


FionnD:
hi @aga11313 @alexcg1 @theUnkownName 

Do we need both the table and bullet points in the overview section?

aga11313:
> hi @aga11313 @alexcg1 @theUnkownName 
> 
> Do we need both the table and bullet points in the overview section?
> 
> Screenshot 2021-05-12 at 16 54 17
> 

Hm the table is more very specific info? And the prose section is more like a story telling information. I would keep them both personally just keep the second part more brief

example_template.md
```diff
@@ -1,4 +1,10 @@
 # Run the EXAMPLE NAME
+*You can also include a gif with a full demo of the example*
+
+
+ *ADD A TABLE OF CONTENTS HERE *
```
alexcg1, line 5: I think examples already used an automated TOC creator, no? I think it was part of CI/CD. Just need to put the right tags in the code.

I nabbed this from our ancient South Park example:

```markdown
<!-- START doctoc generated TOC please keep comment here to allow auto update -->
<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
**Table of Contents**

- [Overview](#overview)
- [Try out this demo by yourself](#try-out-this-demo-by-yourself)
- [Prerequirements](#prerequirements)
- [Prepare the data](#prepare-the-data)
- [Define the Flows](#define-the-flows)
- [Run the Flows](#run-the-flows)
- [Dive into the Pods](#dive-into-the-pods)
- [Wrap up](#wrap-up)
- [Next Steps](#next-steps)
- [Documentation](#documentation)
- [Stay tuned](#stay-tuned)
- [License](#license)

<!-- END doctoc generated TOC please keep comment here to allow auto update -->
```

Stuff between tags is auto-generated each time. I assume it should still work for current examples

FionnD: I always assumed you had to run doctoc locally? I don't see it mentioned anywhere in our github workflow?

theUnkownName:
> I would like to keep it short, and prevent text heavy. But I'm not to pushed about it.
> 
> > hi @aga11313 @alexcg1 @theUnkownName
> > Do we need both the table and bullet points in the overview section?
> > Screenshot 2021-05-12 at 16 54 17
> 
> Hm the table is more very specific info? And the prose section is more like a story telling information. I would keep them both personally just keep the second part more brief

I agree with Agnes. I would vote to keep only the table. It explains it better and we shouldn't be so verbose with the bullet points.

FionnD:
> > I would like to keep it short, and prevent text heavy. But I'm not to pushed about it.
> > > hi @aga11313 @alexcg1 @theUnkownName
> > > Do we need both the table and bullet points in the overview section?
> > > Screenshot 2021-05-12 at 16 54 17
> > 
> > 
> > Hm the table is more very specific info? And the prose section is more like a story telling information. I would keep them both personally just keep the second part more brief
> 
> I agree with Agnes. I would vote to keep only the table. It explains it better and we shouldn't be so verbose with the bullet points.

So only the table?

example_template.md
alexcg1: Ah, maybe? I guess I didn't often change the headings when I edited examples so the issue never arose for me.

Want to run a quick test? Try replacing your TOC with the empty doctoc tags and see what happens? Can always revert. if it works, saves a lot of time of time for everyone in future. If not, just revert and it's like 2 minutes of wasted effort

theUnkownName:
I would say only the table and not the bullet points, yes

example_template.md
FionnD: Tested and it didn't work! You have to run it locally. So I would leave it in the template and require people to manually make sure it works.

FionnD:
Update: 
1. I've add a TOC to remind people to add it. 
2. I've remove the paragraph under the table to streamline the template.

theUnkownName (approved):

PR #616, opened 2021-05-18
Added commands to speed up cloning using sparse checkout

gitcommitshow:
Some people reported that cloning examples take time, so now I have added quick commands to copy only one example instead of all of them

### DIFF
--- a/multimodal-search-pdf/README.md
+++ b/multimodal-search-pdf/README.md
@@ -37,6 +37,18 @@ What's included in this example:
   * [Community](#community)
   * [License](#license)
     
+<details>
+ <summary>How to clone this repo❓</summary>
+ 
+ Because this repo contains many [more example projects](https://github.com/jina-ai/examples/), if you want to clone only this example, you can do this as following
+ 
+ ```bash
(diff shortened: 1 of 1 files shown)


multimodal-search-pdf/README.md
```diff
@@ -37,6 +37,18 @@ What's included in this example:
...
   * [License](#license)
     
+<details>
+ <summary>How to clone this repo❓</summary>
+ 
+ Because this repo contains many [more example projects](https://github.com/jina-ai/examples/), if you want to clone only this example, you can do this as following
+ 
+ ```bash
+ git clone --depth 1 --filter=blob:none --sparse https://github.com/jina-ai/examples
+ git sparse-checkout set multimodal-search-pdf
```
nan-wang, line 47:
```suggestion
cd examples
git sparse-checkout set multimodal-search-pdf
```

nan-wang (changes requested):
Good job!

multimodal-search-pdf/README.md
gitcommitshow: I have added the `cd examples` now in the latest commit

nan-wang (approved):

PR #617, opened 2021-05-18
chore: update jinacld version

slettner:
Use the new version of `jinacld_tools` (used for pulling datasets from S3).

### DIFF
--- a/util/pull_dataset.py
+++ b/util/pull_dataset.py
@@ -40,7 +40,7 @@ def main(data_set: str, pull_to_dir: str):
     _check_credentials_exist()
     assert os.path.isdir(pull_to_dir), "The pull dir parameter must be an existing directory"
     save_path = os.path.join(pull_to_dir, data_set)
-    s3 = S3(BUCKET_NAME)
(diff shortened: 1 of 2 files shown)


jacobowitz (dismissed):

util/pull_dataset.py
```diff
@@ -3,7 +3,7 @@
 import os
 
 import click
-from jinacld_tools.aws.services.s3 import S3
+from jinacld_tools.aws.services.s3 import S3Bucket
```
cristianmtr, line 6: Won't this fail, since it's expecting `S3` below, at https://github.com/jina-ai/examples/blob/chore-update-cloud-helper/util/pull_dataset.py#L43

cristianmtr (changes requested):
I think you need to change the variable name

util/pull_dataset.py
slettner: Oh shoot, thank Cristian.

slettner: 257e54f6e340c6b9f4eb1d4d530bfa91b0e76725

cristianmtr (approved):

PR #623, opened 2021-05-20
update cross modal search to 1.2

nan-wang:
update cross modal search to 1.2

### DIFF
--- a/cross-modal-search/Dockerfile
+++ b/cross-modal-search/Dockerfile
@@ -1,4 +1,4 @@
-FROM jinaai/jina:1.0.8
+FROM jinaai/jina:1.2.0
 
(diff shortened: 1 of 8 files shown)


nan-wang:
The `image_encoder` is broken with the following error message,
```output
RuntimeError: Given groups=1, weight of size [768, 3, 32, 32], expected input[1, 224, 224, 3] to have 3 channels, but got 224 channels instead
```

bwanglzu (approved):
LGTM!

bwanglzu:
testing with @davidbp seems it's not working

bwanglzu:
jina version is not identical to the encoder version, so query hangs.

PR #625, opened 2021-05-20
fix the ci using git instead of pypi

nan-wang:
fix the ci using git instead of pypi

### DIFF
--- a/advanced-vector-search/requirements.txt
+++ b/advanced-vector-search/requirements.txt
@@ -1,4 +1,4 @@
+git+https://github.com/jina-ai/jina.git@v1.2.0#egg=jina[hub, http]
 docker==4.0.1
 click==7.1.2
-jina[hub, http]==1.2.0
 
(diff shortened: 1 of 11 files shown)


alexcg1:
I just tried running `pip install git+https://github.com/jina-ai/jina.git@v1.2.0#egg=jina[hub, http]` and it outputs:

```
ERROR: Invalid requirement: 'jina[hub,'                                                      
WARNING: You are using pip version 21.0.1; however, version 21.1.1 is available.             
You should consider upgrading via the '/home/alexcg/tmp/ci-test/env/bin/python -m pip install --upgrade pip' command.
```

We need to change all occurrences to remove the space after the comma:

#### Before

```
git+https://github.com/jina-ai/jina.git@v1.2.0#egg=jina[hub, http]
```

#### After

```
git+https://github.com/jina-ai/jina.git@v1.2.0#egg=jina[hub,http]
```

alexcg1:
[Another option](https://packaging.python.org/guides/hosting-your-own-index/) instead of going via git

alexcg1:
I just realized that this is great for filtering out our own CI pipeline. But it also filters out every "real" user who wants to play with Jina examples. They need to be counted.

Thought: How about we have TWO requirements files?

- `requirements.txt` - revert back to using `pip` - this is how user would use the example
- `tests/requirements.txt` - use `git` - this is how CI system will test the example

That way we count real users but don't count the CI.

@nan-wang thoughts?

PR #632, opened 2021-06-01
fix pip numbers leakage

alexcg1:
Fixes https://github.com/jina-ai/internal-tasks/issues/86

At present, anything installed from `requirements.txt` in examples repo is not tracked by pypi, so we can't get those metrics. But do we even want those numbers?

- If it's our CI doing that, we DON'T want those numbers. 
- If it's our users, we DO want those numbers. 
 
This PR creates a separate requirements.txt in each example's `test/` directory, and tells our CI to install from that. 

It reverts `./requirements.txt` to installing from pypi, which is what real users would do.

- test: separate reqs for users and ci. users install direct
- test: set ci to install requirements.txt from test dir

### DIFF
--- a/advanced-vector-search/requirements.txt
+++ b/advanced-vector-search/requirements.txt
@@ -1,4 +1,4 @@
-git+https://github.com/jina-ai/jina.git@v1.2.0#egg=jina[hub, http]
+jina[hub, http]
 docker==4.0.1
 click==7.1.2
 
--- a/advanced-vector-search/requirements.txt
+++ b/advanced-vector-search/requirements.txt
@@ -1,4 +1,4 @@
-jina[hub, http]
+jina[hub, http]==1.2.0
 docker==4.0.1
 click==7.1.2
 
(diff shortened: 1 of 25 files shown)


alexcg1:
Thank god for vim's automation via `.`. That saved me a lot of work!

advanced-vector-search/requirements.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/jina-ai/jina.git@v1.2.0#egg=jina[hub, http]
+jina[hub, http]
```
nan-wang, line 1:
```suggestion
jina[hub, http]==1.2.0
```

nan-wang (changes requested):
We need to fix the version

advanced-vector-search/requirements.txt
alexcg1: Fixed for all

nan-wang (approved):
LGTM👍

PR #645, opened 2021-06-07
refactor(nlp-simple-template): kill with fire

alexcg1:
This doesn't include anything beyond that of an old version of the wiki example. Not even worth putting in the legacy repo. Terminating with extreme prejudice.


cristianmtr (commented):
I like deleting stuff

cristianmtr (approved):

alexcg1:
Nothing brings my heart more joy than going on a deleting spree. If we have no code then we can't have bugs!

FionnD (approved):

PR #662, opened 2021-06-29
fix: fix error

Yongxuanzhang:
I remember we will not keep this example after 2.0 official release right? @slettner


slettner:
The example will go to legacy

slettner (approved):

PR #670, opened 2021-07-07
fix: bump Jina version and Port ImageNormalizer and CLIPImageEncoder to use jinahub+docker

winstonww:
This PR bumps the Jina version of cross-modal-search to 2.0.4 and port ImageNormalizer and CLIPImageEncoder to use jinahub+docker

### DIFF
--- a/cross-modal-search/Dockerfile
+++ b/cross-modal-search/Dockerfile
@@ -1,4 +1,4 @@
-FROM jinaai/jina:1.2.0
+FROM jinaai/jina:2.0.8-perf
 
 RUN apt-get -y update && apt-get install -y git wget gcc unzip libmagic1
 
--- a/cross-modal-search/requirements.txt
+++ b/cross-modal-search/requirements.txt
@@ -1,7 +1,7 @@
-jina[hub, http]~=2.0.0
+jina[hub, http, chatbot]~=2.0.0
 nltk==3.5
 pycocotools==2.0.2
 python-magic==0.4.20
 click==7.1.2
 kaggle==1.5.12
-git+https://github.com/openai/CLIP.git
+git+git://github.com/jina-ai/jina-commons
--- a/cross-modal-search/flows/flow-index.yml
+++ b/cross-modal-search/flows/flow-index.yml
@@ -6,15 +6,13 @@ with:
   port_expose: 45678
   workspace: $JINA_WORKSPACE
 pods:
-  - name: loader                                      # load images from the dataset of image-caption pairs
+  - name: image_normalizer                                  # normalize the dimension of the images
+    uses: pods/image-normalize.yml
     uses: 'jinahub+docker://ImageNormalizer'
+    override_with: {'dtype': 'uint8'}
     shards: $JINA_PARALLEL
     read_only: true
     needs: [gateway]
-  - name: image_normalizer                                  # normalize the dimension of the images
-    uses: pods/image-normalize.yml
-    shards: $JINA_PARALLEL
-    read_only: true
   - name: image_encoder                               # encode images into embeddings with CLIP model
     uses: 'jinahub+docker://CLIPImageEncoder'
     volumes: $HOME/.cache/clip:/root/.cache/clip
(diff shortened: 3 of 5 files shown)


cross-modal-search/requirements.txt
```diff
@@ -1,7 +1,7 @@
-jina[hub, http]==2.0.0rc5
+jina[hub, http]==2.0.4
```
mapleeit, line 1: This better be `~=2.0.0`

numb3r3:
```suggestion
jina[hub, http]~=2.0.0
```

numb3r3: test

numb3r3: test2

numb3r3 (changes requested):

cross-modal-search/requirements.txt
winstonww: Addressed, please see [0e3acad](https://github.com/jina-ai/examples/pull/670/commits/0e3acad28fc311ec836f65e99209d5f1b84de393) for the change.

winstonww:
The current PR depends on this [PR](https://github.com/jina-ai/executor-image-normalizer/pull/15). Need to pass `uint8` dtype to ImageNormalizer so that CLIPImageEncoder receives the expected input.

nan-wang:
> The current PR depends on this PR. Need to pass uint8 dtype to ImageNormalizer so that CLIPImageEncoder receives the expected input.

You can convert it to `uint8` when converting image into `blob`. Check out the codes at 
https://github.com/jina-ai/eap-coohom/blob/main/pods/executors.py#L121
https://github.com/jina-ai/eap-coohom/blob/main/pods/executors.py#L24

winstonww:
Thank you for the suggestion. I was wondering if it's possible to just pass `target_dtype` into `ImageNormalizer` with `numpy.uint8` as I believe this also works after the latest [commit](https://github.com/jina-ai/examples/pull/670/commits/eecaa80adf0b516a120cef03ee557712c163b0a4)? I recently made the change to merge the `ImageReader` logic (i.e. converting image to `blob`) into `ImageNormalizer` by creating the function `_convert_image_to_blob` in `ImageNormalizer`. Do I understand correctly that we also want to add the line (or anything that resembles the line) ` _blob = np.array(_blob, dtype='uint8')` in `ImageNormalizer`'s `_convert_image_to_blob`?

nan-wang:
> Thank you for the suggestion. I was wondering if it's possible to just pass target_dtype into ImageNormalizer with numpy.uint8 as I believe this also works after the latest commit? I recently made the change to merge the ImageReader logic (i.e. converting image to blob) into ImageNormalizer by creating the function _convert_image_to_blob in ImageNormalizer. Do I understand correctly that we also want to add the line (or anything that resembles the line) _blob = np.array(_blob, dtype='uint8') in ImageNormalizer's _convert_image_to_blob?

@winstonww We don't need normalizer anymore. The `CLIPImageEncoder` can accept images of arbitrary sizes as long as it is `uint8`. Please refer more details to https://github.com/jina-ai/executor-clip-image#inputs

winstonww:
As discussed with @numb3r3 , currently cross-modal-search example does not work well with docker. We suspect that the issue has to do with flow having both CLIPImageEncoder and CLIPTextEncoder in it, but this needs to be investigated. Nevertheless, the current cross-modal-search example runs successfully in local environment.

cross-modal-search/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM jinaai/jina:1.2.0
+FROM jinaai/jina:2.0.8-perf
```
winstonww, line 1: @Yongxuanzhang  Version is now bumped to 2.0.8. However, we are seeing an issue happening when running cross-modal-search with docker, detailed [here](https://github.com/jina-ai/examples/pull/670#issuecomment-883190942). This needs to be further investigated. Thanks.

numb3r3 (approved):

PR #673, opened 2021-07-09
chore: move unsupported examples to legacy repo

alexcg1:
Will come back to Wikipedia since it may be a dependent of query-while-indexing


nan-wang (approved):
LGTM👍

PR #704, opened 2021-08-24
Feat audio to audio search

winstonww:
Following needs to be done:

- [ ] Add tests
- [ ] Update READMEs
- [ ] Polish code


winstonww:
Tests are now added but are failing on Github - currently under investigation. Also doing experiments to collect accuracy results using various segmenters/encoders.

PR #709, opened 2021-08-27
feat: enhance AudioCLIPEncoder support

winstonww:
This PR:
~~- [x] adds a local `AudioCLIPCrafter` executor to adjust sample rate of audio before passing the blob to `AudioCLIPEncoder` for processing~~
- [x] Add `clip` option to `test_audiio_to_audio_search.py`

Thanks.


nan-wang (approved):
LGTM👍

nan-wang:
> Migrating this logic to AudioCLIPEncoder will safeguard users from providing audios with wrong sample rate.
Good point!

PR #718, opened 2021-09-17
chore(qwi): update to 2.1.0

deepankarm:
Instead of using `master-daemon` image for JinaD, this PR changes both jina version & jinad docker image version to `2.1.0`.


cristianmtr (approved):

cristianmtr:
Did this come out of the Slack question we got some days ago from a user?

deepankarm:
> Did this come out of the Slack question we got some days ago from a user?

That seems to be a different issue. This example works even without changing the versions. This PR just makes sure the same version is used across.

PR #725, opened 2021-09-21
fix: multires usage of simple indexer

cristianmtr:


bwanglzu (approved):
LGTM, but what is `old`?

cristianmtr:
> LGTM, but what is `old`?

The version of SimpleIndexer that works with this example. We'd need to adapt them to the new API at some point though still.

bwanglzu (approved):
LGTM!

